PR #23, opened 2024-07-05
Add index.yaml

DamienDeepgram:
This enables


DamienDeepgram:
Found it here: https://github.com/deepgram/self-hosted-resources/blob/gh-pages/index.yaml

PR #26, opened 2024-07-12
CI and PR template improvements

bd-g:
## Proposed changes

This change improves the DevOps portion of this repo in the following ways:

- Adds a standardized Pull Request template
- Adds a CI check to ensure that the Helm chart appVersion is not updated without also bumping the chart Version
- Adds a CI check to ensure that the chart changelog is updated if the chart is updated
- Improves reliability of the publish GH action
  - The `chart-releaser-action` will attempt to run if git detects any changes at all in the chart directory. 
  - This means if a change was merged, but we aren't yet cutting a release by bumping the version number, git will still detect changes and so `chart-releaser-action` will attempt to run.
  - It will create a tarball and attempt to create a release, but under the same version as the current/latest version. GitHub will return an error complaining that a release already exists under that name/version:
   ```
   Error: error creating GitHub release deepgram-self-hosted-0.2.2-beta: POST https://api.github.com/repos/deepgram/self-hosted-resources/releases: 422 Validation Failed [{Resource:Release Field:tag_name Code:already_exists Message:}]
  ```
  - So instead, this PR adds an explicit check to see if the Chart version has been bumped before taking the GH action

## Types of changes

## Checklist

  - I tested all GH action modifications locally using [act](https://github.com/nektos/act).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,3 +38,25 @@ jobs:
             echo "README.md is out of date. Please run helm-docs and commit the changes."
             exit 1
           fi
+
+      - name: Check that chart version is bumped if chart appVersion is bumped
+        run: |
+          CHART_DIR="charts/deepgram-self-hosted"
+          if [ ! -f "$CHART_DIR/Chart.yaml" ]; then
+            echo "Error: Chart.yaml not found in $CHART_DIR"
+            exit 1
+          fi
+
+          OLD_VERSION=$(git show $GITHUB_BASE_REF:$CHART_DIR/Chart.yaml | grep '^version:' | awk '{print $2}')
+          NEW_VERSION=$(grep '^version:' $CHART_DIR/Chart.yaml | awk '{print $2}')
+          OLD_APP_VERSION=$(git show $GITHUB_BASE_REF:$CHART_DIR/Chart.yaml | grep '^appVersion:' | awk '{print $2}')
+          NEW_APP_VERSION=$(grep '^appVersion:' $CHART_DIR/Chart.yaml | awk '{print $2}')
+
+          if [[ "$OLD_VERSION" = "$NEW_VERSION" ]] && [[ "$OLD_APP_VERSION" != "$NEW_APP_VERSION" ]]; then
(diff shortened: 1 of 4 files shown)


bd-g:
Need to add one more commit here to add a simple CI check that the chart CHANGELOG.md is updated whenever changes are made to the chart.

.github/workflows/ci.yml
```diff
@@ -38,3 +38,25 @@ jobs:
...
+          NEW_VERSION=$(grep '^version:' $CHART_DIR/Chart.yaml | awk '{print $2}')
+          OLD_APP_VERSION=$(git show $GITHUB_BASE_REF:$CHART_DIR/Chart.yaml | grep '^appVersion:' | awk '{print $2}')
+          NEW_APP_VERSION=$(grep '^appVersion:' $CHART_DIR/Chart.yaml | awk '{print $2}')
+
+          if [[ "$OLD_VERSION" = "$NEW_VERSION" ]] && [[ "$OLD_APP_VERSION" != "$NEW_APP_VERSION" ]]; then
+            echo "Error: In $CHART_DIR/Chart.yaml, appVersion has been changed but version has not been bumped"
+            echo "Old appVersion: $OLD_APP_VERSION"
+            echo "New appVersion: $NEW_APP_VERSION"
+            echo "Chart version: $NEW_VERSION"
+            echo "Chart version should be bumped anytime the application version is bumped."
```
jkroll-deepgram, line 60: Is this bump done manually, or through helm-docs?

bd-g: Helm-docs just updates the README based on the contents of the Chart.yaml. Ultimately, the source of truth for the version is going to be the Chart.yaml, which needs to be manually edited. 

Update scenarios for these versions could include:
- No version bump. We are making changes that we don't want to release yet, but want to merge. 
- Bump only the chart version. This could include an update to the templates that doesn't bump the container tags (which map to the Deepgram self hosted release and is what we use for the appVersion)
- Bump both the chart and app version for an update that includes changing the container tags. 

Any other combination of version bump, like updating the app version but not chart version, shouldn't be allowed.

jkroll-deepgram (approved):

PR #28, opened 2024-07-12
[ISSUE-27] Fix EFS Storage Class Mismatch with Name Prefix

paulliwog:
This change fixes a template issue when using AWS EFS as the storage for self-hosted deployments

### DIFF
--- a/charts/deepgram-self-hosted/CHANGELOG.md
+++ b/charts/deepgram-self-hosted/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this Helm chart will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+
+- Resolves a mismatch between PVC and SC prefix naming convention.
+
 ## [0.2.2-beta] - 2024-06-27
 
 ### Added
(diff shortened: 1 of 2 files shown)


bd-g (changes requested):
LGTM! This shouldn't affect functionality, but it does keep consistent naming across Deepgram resources. 

Could you add a line to the CHANGELOG.md under a new header called "Unreleased" describing the change in the specified format there?

johnson-rl (approved):

paulliwog:
> LGTM! This shouldn't affect functionality, but it does keep consistent naming across Deepgram resources.
> 
> Could you add a line to the CHANGELOG.md under a new header called "Unreleased" describing the change in the specified format there?

I'll get that added.  This was impacting functionality in our cluster as we are using Karpenter Autoscaler; the name mismatch prevented karpenter from scheduling the engine pod, so it wouldn't spin up a new node for it.

bd-g:
That's good to know! This hasn't been explicitly tested with Karpenter, so I wasn't aware of that. Thanks for the heads up!

In that case, if you have autoscaling enabled to include the HPA manifests, make sure to explicitly disable the cluster-autoscaler dependency.

bd-g:
Thank you! I'll cut a release for this later today. 

Thanks for your patience on the Changelog request. The other open PR in this report will include a few more CI checks to help guide PR submissions to include that 😀

bd-g (approved):
LGTM!

PR #43, opened 2024-09-27
Add container orchestrator environment variable

bd-g:
## Proposed changes

Add an environment variable to container launch templates to enable more detailed support for self-hosted deployments.

## Types of changes

## Checklist

### DIFF
--- a/docker/docker-compose.license-proxy.yml
+++ b/docker/docker-compose.license-proxy.yml
@@ -1,5 +1,11 @@
 # Make sure to replace placeholder paths to config files and model directories
 
+x-env: &env
+  environment:
+    # Make sure you `export` your self-hosted API key secret in your local environment
+    DEEPGRAM_API_KEY: "${DEEPGRAM_API_KEY}"
+    DEEPGRAM_DEPLOYMENT_ORCHESTRATOR: "docker-compose"
+
 services:
   # The speech API service.
   api:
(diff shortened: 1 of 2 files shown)


charts/deepgram-self-hosted/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 ### Added
 
 - Automatic model management on AWS now supports model removal. See the `engine.modelManager.models.remove` section in the `values.yaml` file for details.
+- Container orchestrator environment variable added to improve support.
```
jkroll-deepgram, line 12: Worth calling out the 3 supported values, since it's a small finite set and we could potentially add/change it in the future?

docker/docker-compose.license-proxy.yml
```diff
@@ -14,6 +14,7 @@ services:
     # Make sure you `export` your self-hosted API key secret in your local environment
     environment:
       DEEPGRAM_API_KEY: "${DEEPGRAM_API_KEY}"
+      DEEPGRAM_DEPLOYMENT_ORCHESTRATOR: "docker-compose"
```
jkroll-deepgram, line 17: Any way to make this a static variable or enum of some type, or is hard-coding the same string multiple times the best option on these?

charts/deepgram-self-hosted/CHANGELOG.md
bd-g: It's not necessarily intended to be change or configured - I don't think it needs to be called out.

Also, the cardinality will increase, as the Helm value will change for each version (but the prefix will stay the same).

docker/docker-compose.license-proxy.yml
bd-g: I wasn't previously aware, which is why there is duplication in other places, but your question prompted me to go looking, and I found Docker Compose extension fields. Let me see if they're also compatible with Podman.

bd-g: The Compose protocol is fairly limited. I was able to reduce the duplication of the API key and the common orchestrator variable. I tried to do the same for the release tag on each image, but I can edit only a subsection of a field.

It's also not supported by `podman-compose` right now, so can only use this for the Docker Compose templates.

jkroll-deepgram (approved):

PR #51, opened 2024-12-02
Remove resource limits on Compose file templates

bd-g:
## Proposed changes

Resource limits can be an effective way to prevent containers from starving eachother. 

However, in a single VM running multiple Deepgram containers (most common use case with Docker), allowing containers to have more ebb and flow in their resource usage can lead to greater performance and easier management. 

This also prevents artificially low limits being enforced when these templates are used for more powerful instance types, but users forget to increase the default limits.

## Types of changes

  - Updated best practice on Compose templates 

## Checklist

### DIFF
--- a/docker/docker-compose.license-proxy.yml
+++ b/docker/docker-compose.license-proxy.yml
@@ -28,15 +28,6 @@ services:
     # Invoke the API server
     command: -v serve /api.toml
 
-    # Default resource limits. Increase as needed.
-    # For more details, see:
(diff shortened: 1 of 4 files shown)


jkroll-deepgram:
Makes sense. Do we want to leave it commented-out, or document elsewhere how/when to impose resource limits? Or is that considered outside the scope of what we want to educate here?

jkroll-deepgram (approved):

PR #60, opened 2025-03-06
Voice Agent support

bd-g:
## Proposed changes

Add self-hosted Voice Agent support to the bundled Helm chart. Voice Agent will not initially be supported by default templates for Docker/Podman.

## Types of changes

## Checklist

  - Please describe your testing setup and methodology here


charts/deepgram-self-hosted/templates/engine/engine.deployment.yaml
```diff
@@ -1,76 +1,90 @@
...
   selector:
     matchLabels:
       app: deepgram-engine
-      {{ include "deepgram-self-hosted.selectorLabels" . }}
-  replicas: {{ .Values.scaling.replicas.engine }}
+      {{- if $type }}
+      engine-type: {{ $type }}
+      {{- end }}
+      {{ include "deepgram-self-hosted.selectorLabels" $ }}
+  replicas: {{ $.Values.scaling.replicas.engine }}
```
bd-g, line 34: Speech-to-text, text-to-speech, and end-of-turn will scale very differently. Once load goes above the connections supported by just 3 Engines (1 per type), we'll want to support independently scaling the three engine types. I would recommend the following structure:

values.yaml
```
## When using standard
scaling:
  replicas:
    api: 1
    engine: 1
## Override in samples yaml when using Voice Agent
scaling:
  replicas:
    api: 1
    engine: 
       agent-speech-to-text: 1
       agent-text-to-speech: 1
       agent-end-of-turn: 1
```

in this file, when using standard:
```
 replicas: {{ if kindIs "map" .Values.scaling.replicas.engine }}{{ required "Expected integer for Engine replicas, got a map" nil }}{{ else }}{{ .Values.scaling.replicas.engine }}{{ end }}
```

in this file, when using voice agent:
```
  replicas: {{ if kindIs "map" $.Values.scaling.replicas.engine }}
    {{- $replicaCount := index $.Values.scaling.replicas.engine (printf "agent-%s" $type) | default 1 }}
    {{ $replicaCount }}
  {{- else }}
    {{- required (printf "Expected map for Voice Agent Engine replicas, got %s" (kindOf $.Values.scaling.replicas.engine)) nil }}
  {{- end }}
```

charts/deepgram-self-hosted/templates/engine/engine.hpa.yaml
```diff
@@ -1,4 +1,8 @@
-{{- if .Values.scaling.auto.enabled -}}
+{{- if and .Values.scaling.auto.enabled .Values.agent.enabled }}
+{{- fail "Error: Autoscaling is not yet supported for voice agent" }}
+{{- end }}
+{{- if .Values.scaling.auto.enabled }}
```
bd-g, line 4: Nice clear error message.

charts/deepgram-self-hosted/templates/NOTES.txt
```diff
@@ -1,4 +1,4 @@
-This is an beta Helm chart for Deepgram self-hosted deployments.
+This is a Helm chart for Deepgram self-hosted deployments.
```
bd-g, line 1: great catch

charts/deepgram-self-hosted/templates/api/api.deployment.yaml
```diff
@@ -50,9 +50,57 @@ spec:
...
+          value: ""
+        {{- end }}
+        {{- if not .Values.global.cartesiaSecretRef }}
+        - name: CARTESIA_API_KEY
+          value: ""
+        {{- end }}
+        {{- if not .Values.global.xaiSecretRef }}
+        - name: XAI_API_KEY
+          value: ""
+        {{- end }}
```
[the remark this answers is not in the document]
jkroll-deepgram: Pestering resumed :)

charts/deepgram-self-hosted/templates/engine/engine.deployment.yaml
jkroll-deepgram: https://github.com/deepgram/self-hosted-resources/pull/60/commits/d4aa74eaca3dbf6425f37124859ab0599fc64bf4

charts/deepgram-self-hosted/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v2
 name: deepgram-self-hosted
 type: application
-version: 0.11.0
-appVersion: "release-250307"
+version: 0.12.0
+appVersion: "release-250307-voice-agent"
```
jkroll-deepgram, line 5: @bd-g what do you think is best to do here in version handling? I'd prefer not to do something like `release-250318` because that doesn't align with a container image release, and I don't want to leave it as `release-250307` because I'm changing the template from the original one put out at the same time of the self-hosted release earlier in march. But this `release-250307-voice-agent` version format is non-standard.

bd-g: Actually, it's very idiomatic here to bump the `version` but leave the `appVersion` as release-250307. 

The two fields are distinct for a purpose - version is for the configuration itself (think, the Helm chart and its values), whereas appVersion is specifically to track the version of the underlying application being managed - in our case, the API, Engine, and License Proxy containers. 

So we're actually perfectly set-up for a change like this. The appVersion stays the same, and we bump the Chart version to indicate the new configuration support for Voice Agent.

charts/deepgram-self-hosted/samples/04-voice-agent-aws.values.yaml
```diff
@@ -0,0 +1,164 @@
...
+          batch:
+            requestsPerPod: # Discuss a reasonable value with your Deepgram Account Representative
+          streaming:
+            requestsPerPod: # Discuss a reasonable value with your Deepgram Account Representative
+        textToSpeech:
+          batch:
+            requestsPerPod: # Discuss a reasonable value with your Deepgram Account Representative
+        # Discuss a reasoanble value with your Deepgram Account Representative
+        # Must also set engine.concurrencyLimit.activeRequests if using request ratio for autoscaling
+        requestCapacityRatio:
```
bd-g, line 61: Should probably remove this section from the `samples.yaml` entirely, as the default is to disable it, and auto-scaling isn't supported for VA yet.

charts/deepgram-self-hosted/samples/04-voice-agent-aws.values.yaml
```diff
@@ -0,0 +1,164 @@
...
+          fileSystemId: fs-xxxxxxxxxxxxxxxx # Replace with your EFS ID
+    models:
+      add:
+        # Recommended Voice Agent models:
+        # - Nova-3 non-formatted (nova-3-general.en.non-formatted.streaming.bf05427e.dg)
+        # - EOT (end-of-turn.4e9dbd96.dg)
+        # - Phonemizer for TTS (phonemizer.en.d6b10ae6.dg)
+        # - Aura-2 TTS voice model(s) (e.g. aura-asteria-en.ecb76e9d.dg)
+        # - Any other streaming STT models of choice (e.g. Nova-2 instead of Nova-3)
+        # - https://link-to-model-1.dg
```
bd-g, line 123: I'd suggest using a wildcard instead of a specific UUID on these recommendations, as model versions may change, and the intent is to communicate "use a Nova-3 non-formatted streaming model", not "use *this specific* Nova-3 non-formatted streaming model".

charts/deepgram-self-hosted/samples/04-voice-agent-aws.values.yaml
```diff
@@ -0,0 +1,164 @@
...
+        # ...
+      remove:
+        # - https://link-to-old-model-1.dg # Replace these with identifiers for any models already present
+        # - https://link-to-old-model-2.dg #   in the EFS that you'd like removed. For a new installation,
+        # - name-of-old-model-3.dg #   this will likely be empty.
```
bd-g, line 128:
```suggestion
        # - https://link-to-old-model-1.dg # Replace these with identifiers for any models already present
        # - https://link-to-old-model-2.dg #   in the EFS that you'd like removed. For a new installation,
        # - name-of-old-model-3.dg          #   this will likely be empty.
```

charts/deepgram-self-hosted/templates/engine/engine.deployment.yaml
```diff
@@ -1,76 +1,103 @@
...
+    {{- else }}
+    {{- required (printf "Expected map for Voice Agent Engine replicas, got %s" (kindOf $.Values.scaling.replicas.engine)) nil }}
+    {{- end }}
+  {{- else }}
+    {{- if kindIs "map" .Values.scaling.replicas.engine }}
+    {{- required "Expected integer for Engine replicas, got a map" nil }}
+    {{- else }}
+    {{- .Values.scaling.replicas.engine }}
+    {{- end }}
+  {{- end }}
```
bd-g, line 47: It looks like moved to using the map for Voice Agent replica counts, but the VA `samples.yaml` still have a plain `int` for `scaling.replicas.engine`. Do we need to align those two?

charts/deepgram-self-hosted/values.yaml
```diff
@@ -99,6 +106,78 @@ scaling:
               value: 25
               periodSeconds: 60
 
+agent:
```
bd-g, line 109: We should add documentation to these values, similar to documentation you can see throughout the file. The pattern is:

```
# -- {string describing the value here}
```

If you are describing a top-level object that has nesting under it, we don't want the default to show in the generated docs (it gets very large). So you'll want to add the following as well:

```
# @default -- ``
```

See the various examples in this file, and then how the documentation is generated in the README.

jkroll-deepgram: https://github.com/deepgram/self-hosted-resources/pull/60/commits/a7d4ea02ba7fb211559f86d85a7c526ecfe5baac

charts/deepgram-self-hosted/Chart.yaml
jkroll-deepgram: https://github.com/deepgram/self-hosted-resources/pull/60/commits/df76e809ec55d9c5f2967b9429ec889cb5085d2f

charts/deepgram-self-hosted/samples/04-voice-agent-aws.values.yaml
jkroll-deepgram: https://github.com/deepgram/self-hosted-resources/pull/60/commits/2aa0f539c5bb8a072d88b22fbe9ca7295d30dbb9

charts/deepgram-self-hosted/samples/04-voice-agent-aws.values.yaml
```diff
@@ -0,0 +1,164 @@
...
+        efs:
+          enabled: true
+          fileSystemId: fs-xxxxxxxxxxxxxxxx # Replace with your EFS ID
+    models:
+      add:
+        # Recommended Voice Agent models:
+        # - Nova-3 non-formatted (nova-3-general.en.non-formatted.streaming.bf05427e.dg)
+        # - EOT (end-of-turn.4e9dbd96.dg)
+        # - Phonemizer for TTS (phonemizer.en.d6b10ae6.dg)
+        # - Aura-2 TTS voice model(s) (e.g. aura-asteria-en.ecb76e9d.dg)
```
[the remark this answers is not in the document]
jkroll-deepgram: https://github.com/deepgram/self-hosted-resources/pull/60/commits/34a2213273dc436cfae201a626159d814558a790

charts/deepgram-self-hosted/samples/04-voice-agent-aws.cluster-config.yaml
```diff
@@ -0,0 +1,79 @@
+apiVersion: eksctl.io/v1alpha5
+kind: ClusterConfig
+
+metadata:
+  name: deepgram-self-hosted-cluster-va
```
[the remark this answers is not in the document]
jkroll-deepgram: https://github.com/deepgram/self-hosted-resources/commit/9a4f54fa915c324ff8ccfcfaa56875eabc555720

charts/deepgram-self-hosted/samples/04-voice-agent-aws.values.yaml
```diff
@@ -0,0 +1,241 @@
...
+          name: "Llama 3.1 8B Instruct"
+          tier: "standard"
+          public: false
+    x_ai:
+      name: "xAI"
+      models:
+        grok-2-latest:
+          name: "Grok 2 Latest"
+          tier: "standard"
+          public: false
```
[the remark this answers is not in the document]
jkroll-deepgram: https://github.com/deepgram/self-hosted-resources/commit/963c2b96ebc5601b3d9669f985ae5799de28287a

charts/deepgram-self-hosted/templates/engine/engine.deployment.yaml
jkroll-deepgram: Great catch, thanks. https://github.com/deepgram/self-hosted-resources/commit/c3ae7fdd30c5dc982f751109e6417b63ee0cb862

jkroll-deepgram:
@bd-g I tested this PR again today on EKS:
- First, with the Voice Agent `05-voice-agent-aws.values.yaml` and served self-hosted VA requests (Nova-3 + OpenAI LLM + Aura-1)
- Next, with the basic AWS `01-basic-setup-aws.values.yaml` and served local STT (Nova-3) and TTS (Aura-1) requests

Can you do another review and let me know if you have any outstanding feedback or concerns about me moving forward to merge this PR?

jkroll-deepgram:
Made one more tweak to secrets (https://github.com/deepgram/self-hosted-resources/pull/60/commits/87cb0d0a47b1f6a80fc0884dd44a517280ea85b2) - this worked to correctly consume the OpenAI API key, without raising any errors that the other 3rd party API keys were not set.

charts/deepgram-self-hosted/values.yaml
```diff
@@ -99,6 +101,108 @@ scaling:
...
+  # -- (bool) Whether to enable voice agent. Disabled by default
+  enabled: false
+  # -- (int) Timeout in milliseconds for end-of-turn detection
+  eotTimeoutMs: 3500
+  # -- (int) Maximum number of characters allowed in a conversation history
+  maxConversationChars: 15000
+  # -- Configuration for LLM providers and their available models
+  # @default -- ``
+  llmProviders:
```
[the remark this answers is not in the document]
jkroll-deepgram: I believe as of this change https://github.com/deepgram/stem/pull/2048, now that LLM providers are optional, there is no longer an issue with having these default LLM providers when keys are not present. On older versions, it was a problem.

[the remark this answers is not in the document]
bd-g: ^^ That panic is what was noodling in the back of my brain, and seems like a sharp edge.

PR #106, opened 2025-10-29
Correct hard coded values on Kubernetes resources

brianrudolf:
## Proposed changes

This corrects instances of hard coded values for labels and selectors on various Kubernetes objects. This will allow this chart to be installed multiple times on the same Kubernetes cluster in the same namespace.

Note that the hard coded values are replaced in this change with pre-existing arguments from the `values.yaml` file and match the default value provided by this chart. While this may cause changes for current chart installs that have customized the `namePrefix` values, I don't believe those chart installs would be working correctly as the service selectors would not select the correct pods. Thus, I don't believe there are installed charts that have customized `namePrefix` values and this change should not break anyone's currently installed chart.

I am currently unable to install this chart multiple times (as is advised for using STT, TTS, and Flux as they each require dedicated resources) due to these hardcoded labels and selectors. 

## Types of changes

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

  - I set the following values and was able to install this chart alongside a pre-existing release on Google Cloud Kubernetes cluster 
    - `api.namePrefix`
    - `engine.namePrefix`
    - `licenseProxy.namePrefix`
  - I updated my existing Helm release (which did not customize the `namePrefix` values) with this updated chart and did not see any changes around names, labels, selectors, etc.
- [ ] I have added necessary documentation (if appropriate)

### DIFF
--- a/charts/deepgram-self-hosted/CHANGELOG.md
+++ b/charts/deepgram-self-hosted/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 ### Fixed
 
 - Fixed API templates to use correct `additionalLabels` reference
+- Fixed hard coded labels and selectors
 
 ## [0.22.0] - 2025-10-15
 


pcgeek86 (approved):
LGTM! Thank you for fixing this.

PR #157, opened 2026-04-23
feat(engine): support configurable HTTP readiness probe for engine pods

ayz-lex:
## Summary

- Adds optional `httpGet` configuration for the engine **readiness probe** only
- When `readinessProbe.httpGet.path` is set (e.g., `/v1/status/engine`), the probe uses an HTTP GET request instead of the default TCP socket check
- Startup and liveness probes remain `tcpSocket` (unchanged)
- Defaults to `tcpSocket` (no behavior change for existing deployments)

## Motivation

The Deepgram docs [recommend](https://developers.deepgram.com/docs/self-hosted-status-endpoint) using the `/v1/status/engine` endpoint for health checking, which returns structured status information (Initializing, Ready, Healthy, Critical). The current `tcpSocket` readiness probe only verifies the port is open, not that the engine is in a healthy state to serve inference. Using `httpGet` for the readiness probe allows Kubernetes to remove pods from service when they report unhealthy status, without restarting them.

## Usage

```yaml
engine:
  readinessProbe:
    httpGet:
      path: /v1/status/engine
```

Omitting `httpGet` preserves the existing `tcpSocket` behavior.

## Test plan

- [ ] Verify default behavior (no `httpGet` set) still uses `tcpSocket` — `helm template` output should be unchanged
- [ ] Verify setting `readinessProbe.httpGet.path` switches readiness to HTTP probe
- [ ] Verify startup and liveness probes are unaffected regardless of readiness config

### DIFF
--- a/charts/deepgram-self-hosted/templates/engine/engine.deployment.yaml
+++ b/charts/deepgram-self-hosted/templates/engine/engine.deployment.yaml
@@ -203,8 +203,14 @@ spec:
           periodSeconds: {{ $.Values.engine.livenessProbe.periodSeconds }}
           failureThreshold: {{ $.Values.engine.livenessProbe.failureThreshold }}
         readinessProbe:
+          {{- if $.Values.engine.readinessProbe.httpGet }}
+          httpGet:
+            path: {{ $.Values.engine.readinessProbe.httpGet.path }}
+            port: {{ $.Values.engine.server.port }}
+          {{- else }}
           tcpSocket:
             port: {{ $.Values.engine.server.port }}
+          {{- end }}
           initialDelaySeconds: {{ $.Values.engine.readinessProbe.initialDelaySeconds }}
           periodSeconds: {{ $.Values.engine.readinessProbe.periodSeconds }}
           failureThreshold: {{ $.Values.engine.readinessProbe.failureThreshold }}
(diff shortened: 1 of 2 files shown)


jkroll-deepgram:
@ayz-lex thanks for the chart contribution!

Note that `/v1/status/engine` is served by the API container, not the Engine container ([doc link](https://developers.deepgram.com/docs/metrics-guide#deepgram-api)). Its handler returns 200 OK with the state encoded in the JSON body (`{"engine_connection_status": "Connected" | "Disconnected"}`), depending on whether the API is connected to an Engine or not. The status code is always 200, which is why the chart's existing API readiness probe ([chart link](https://github.com/deepgram/self-hosted-resources/blob/main/charts/deepgram-self-hosted/templates/api/api.deployment.yaml#L186-L194)) uses `exec` + `grep` rather than `httpGet`, which requires non-2xx status codes to function. [Here](https://github.com/deepgram/self-hosted-resources/blob/main/charts/deepgram-self-hosted/templates/engine/engine.deployment.yaml#L205-L210) is the current Engine readiness probe as well.

Are the current API and/or Engine readiness probes not working for you? We're happy to take any other product or chart feedback you're aiming to address with this change.

jkroll-deepgram:
Going to close this out since there's no further updates since [my prior comment](https://github.com/deepgram/self-hosted-resources/pull/157#issuecomment-4339324188).

PR #161, opened 2026-05-04
Upgrade Kubernetes version to 1.35 in AWS cluster configs

jkroll-deepgram:
## Proposed changes

Upgrade Kubernetes version from 1.33 to 1.35 in AWS cluster configs ahead of end-of-support.

See AWS EKS doc: [Understand the Kubernetes version lifecycle on EKS](https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html)

Kubernetes 1.33 (Deepgram chart's prior version):
- Standard support ends July 29, 2026 [in less than 3 months]
- Extended support ends July 29, 2027

Kubernetes 1.35 (Deepgram chart's new version): 
- Standard support ends March 27, 2027
- Extended support ends March 27, 2028

I tested in EKS and it works as expected, no issues updating a Deepgram cluster from 1.33 to 1.35 and serving requests.

## Types of changes

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [ ] I have read the [CONTRIBUTING](../CONTRIBUTING.md) doc
  - Please describe your testing setup and methodology here
- [ ] I have added necessary documentation (if appropriate)

### DIFF
--- a/charts/deepgram-self-hosted/CHANGELOG.md
+++ b/charts/deepgram-self-hosted/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## Unreleased
 
+### Changed
+
+- Updated sample cluster configuration files to use Kubernetes 1.35 (previously 1.33)
+
 ## [0.35.1] - 2026-05-01
 
 ### Changed
--- a/charts/deepgram-self-hosted/samples/01-basic-setup-aws.cluster-config.yaml
+++ b/charts/deepgram-self-hosted/samples/01-basic-setup-aws.cluster-config.yaml
@@ -4,7 +4,7 @@ kind: ClusterConfig
 metadata:
   name: deepgram-self-hosted-cluster
   region: us-west-2
-  version: "1.33"
+  version: "1.35"
 
 iam:
   withOIDC: true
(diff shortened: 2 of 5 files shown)


connerhughes-dg (approved):
LGTM — version bump is safe.

Confirmed this is a docs/samples-only change: the three eksctl sample cluster configs go 1.33 → 1.35, plus the CHANGELOG entry. No Helm template logic is touched.

- EKS support: Verified — EKS has supported Kubernetes 1.35 since Jan 2026 in all regions, and it's in standard support for 14 months. New-cluster creation and upgrades are supported via eksctl, so the copy-paste samples will work as written.
- API compatibility: Audited every apiVersion the chart renders against 1.34/1.35 deprecations/removals. All are long-stable GA APIs, nothing uses a beta or removed API.
- 1.35 gotchas — both clear: The cgroup v1 deprecation doesn't apply since the samples use Ubuntu2204 / AmazonLinux2023 (both cgroup v2 by default). The containerd 2.0 requirement only affects the next upgrade beyond 1.35.

Non-blocking nit: users on an older eksctl will need a recent enough version to recognize version: "1.35" — worth a one-line note in the docs if not already implied, but not a blocker.

PR #167, opened 2026-05-20
Feat/aiworks

nickottocode:
# Add `deepgram-aiworks` Helm chart

New chart at `charts/deepgram-aiworks/` for deploying VoiceWorks
(AIWorks) self-hosted on Kubernetes. Parallel to `deepgram-self-hosted`,
independently versioned, backend-only (no frontend) for v1.

## Summary

- New chart with full template set: Deployment, Service, ConfigMap
  (renders `config.toml` from values), Secret (creds + optional TLS
  material), PVC, plus optional Ingress / HPA / NetworkPolicy /
  ServiceMonitor each gated on its own values flag.
- Helm `helm test` post-install Pod that curls `/api/health` (branches
  on `server.tls.enabled` to use http vs https).
- CI matrix at `.github/workflows/aiworks-chart-ci.yml` — runs
  `helm lint` + `helm template` + `kubeconform` across three values
  scenarios (default, files-mode, ingress + mTLS).
- `publish.yml` extended so version bumps on either `deepgram-self-hosted`
  OR `deepgram-aiworks` trigger a release; handles the "new chart on
  first merge" edge case (no prior `Chart.yaml` in `HEAD^`).
- Chart README with prerequisites, quick start, four example deployments
  (files+open mode, ingress, in-process TLS, postgres), upgrade /
  uninstall recipes, and a values reference table.

## Quick-start example

```sh
helm repo add deepgram https://deepgram.github.io/self-hosted-resources
helm install aiworks deepgram/deepgram-aiworks \
  --set license.apiKey=<deepgram-license-uuid> \
  --set selfHosted.apiKey=<40+-char-shared-secret> \
  --set selfHosted.bootstrapSubject=$(uuidgen | tr A-Z a-z)
```

Connect clients via:
```
ws://<service-or-ingress>/api/repos/<any-uuid>/flows/<flow-slug>/execute
Authorization: Token <selfHosted.apiKey>
```

## Architectural notes for reviewers

- **License-proxy is NOT bundled.** The chart's `license.serverUrl`
  defaults to `http://license-proxy.deepgram.svc.cluster.local:8080`
  with `https://license.deepgram.com` as fallback. Operators reuse the
  license-proxy from an existing `deepgram-self-hosted` install or
  point at a standalone deployment.
- **`${VW_API_KEY}` flows through pod env.** Helm can't inline Secret
  values at template time, so the ConfigMap renders
  `[self_hosted].api_key = "${VW_API_KEY}"` as a literal placeholder.
  Voiceworks's config loader resolves it from `std::env::var` at
  startup — the corresponding backend change shipped in voiceworks
  commit `9302010`. The chart README documents this contract in the
  values reference.
- **`/api/health` is used for all three probes** (readiness / liveness /
  startup). It's unconditionally registered in both `files` and
  `postgres` storage modes, so probe config doesn't need to branch on
  `storage.mode`.
- **Security defaults:** non-root UID 10001, `readOnlyRootFilesystem`,
  all caps dropped. Matches the `voiceworks-self-hosted` image's user.
- **`updateStrategy.rollingUpdate.maxUnavailable: 0`** by default —
  keeps at least one pod serving during rolling upgrades. Long-lived
  WebSocket sessions drain over `terminationGracePeriodSeconds`.

## What's intentionally out of scope

- **Self-hosted frontend** (Next.js editor). Backend-only for v1.
- **Bundled license-proxy.** Operators bring their own (existing
  `deepgram-self-hosted` install or standalone).
- **ACME / Let's Encrypt in-process.** Use an ingress controller for
  public-internet TLS termination.
- **SNI-based multi-cert.** Single cert per listener.

## Companion voiceworks work

This PR is the chart half of a four-plan effort to make AIWorks
production-ready for self-hosted. The matching backend changes ship
on the `feat/self-hosted` branch of github.com/deepgram/voiceworks —
that branch will land separately.

Design spec lives at
`voiceworks/docs/superpowers/specs/2026-05-18-aiworks-self-hosted-production-readiness-design.md`.

## Test plan

- [x] `helm lint` clean against all three CI values files
  (default-values.yaml, files-mode-values.yaml, ingress-mtls-values.yaml)
- [x] `helm template` produces well-formed manifests for each scenario
- [x] CI workflow YAML parses; `kubeconform` step uses
  `-ignore-missing-schemas` to allow CRDs (ServiceMonitor)
- [ ] First merge to `main` triggers chart release via `publish.yml`
  (verify after merge — the "new chart on first merge" branch is
  exercised here for the first time)
- [ ] `helm install` against a real cluster pointing at an existing
  `deepgram-self-hosted` license-proxy
- [ ] `helm test aiworks` returns success in both TLS-on and TLS-off
  configurations

## Required follow-ups (out of this PR)

- voiceworks `feat/self-hosted` branch must land before the chart is
  usable end-to-end (provides the backend image and the
  `${VW_API_KEY}` env-var resolution this chart relies on).
- `developers.deepgram.com` should get an "AIWorks (Self-Hosted)"
  section that links into the chart README and the in-product docs
  shipping in voiceworks.

### DIFF
--- a/.github/workflows/aiworks-chart-ci.yml
+++ b/.github/workflows/aiworks-chart-ci.yml
@@ -1,0 +1,40 @@
+name: aiworks-chart-ci
+
+on:
+  pull_request:
+    paths:
+      - "charts/deepgram-aiworks/**"
+      - ".github/workflows/aiworks-chart-ci.yml"
+  push:
+    branches: [main]
+    paths:
+      - "charts/deepgram-aiworks/**"
+
+jobs:
+  lint-and-template:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        values:
+          - ci/default-values.yaml
+          - ci/files-mode-values.yaml
+          - ci/ingress-mtls-values.yaml
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install helm
+        uses: azure/setup-helm@v4
+
+      - name: helm lint
+        run: helm lint charts/deepgram-aiworks --values charts/deepgram-aiworks/${{ matrix.values }}
+
+      - name: helm template
+        run: helm template aiworks-test charts/deepgram-aiworks --values charts/deepgram-aiworks/${{ matrix.values }} > /tmp/rendered.yaml
+
+      - name: Install kubeconform
+        run: |
+          set -e
+          curl -sL https://github.com/yannh/kubeconform/releases/latest/download/kubeconform-linux-amd64.tar.gz | tar -xz -C /usr/local/bin
+
+      - name: kubeconform render
+        run: kubeconform -strict -summary -kubernetes-version 1.28.0 -ignore-missing-schemas /tmp/rendered.yaml
--- a/charts/deepgram-aiworks/README.md
+++ b/charts/deepgram-aiworks/README.md
@@ -1,0 +1,5 @@
+# deepgram-aiworks
+
+Helm chart for VoiceWorks (AIWorks) self-hosted. See [the design spec](https://github.com/deepgram/voiceworks/blob/master/docs/superpowers/specs/2026-05-18-aiworks-self-hosted-production-readiness-design.md) for context.
+
+Full README content lands in this chart's `0.1.0` final cut — see Plan D Task 6.
--- a/charts/deepgram-aiworks/README.md
+++ b/charts/deepgram-aiworks/README.md
@@ -1,5 +1,203 @@
 # deepgram-aiworks
 
-Helm chart for VoiceWorks (AIWorks) self-hosted. See [the design spec](https://github.com/deepgram/voiceworks/blob/master/docs/superpowers/specs/2026-05-18-aiworks-self-hosted-production-readiness-design.md) for context.
+Helm chart for VoiceWorks (AIWorks) self-hosted. Deploys the AIWorks
+backend alongside or independent of the Deepgram speech distribution
+(`deepgram-self-hosted`). Backend-only for v1; no frontend.
 
-Full README content lands in this chart's `0.1.0` final cut — see Plan D Task 6.
+## Prerequisites
+
+- Kubernetes 1.28+
+- A reachable `license-proxy` endpoint. Two ways to get one:
+  - Already running `deepgram-self-hosted` (most common). Point
+    `license.serverUrl` at `http://license-proxy.<namespace>.svc.cluster.local:8080`.
+  - Standalone license-proxy deployment (out of scope for this chart;
+    see Deepgram self-hosted docs).
+- A self-hosted Deepgram speech endpoint (STT). Point
+  `deepgram.streamingListenUrl` and `deepgram.batchListenUrl` at it.
+- Container registry access to `quay.io/deepgram/voiceworks-self-hosted`.
+  Add an `imagePullSecret` to `imagePullSecrets[]` if needed.
+
+## Quick start
+
+```sh
+helm repo add deepgram https://deepgram.github.io/self-hosted-resources
+helm install aiworks deepgram/deepgram-aiworks \
+  --set license.apiKey=<your-deepgram-api-key-uuid> \
+  --set selfHosted.apiKey=<your-40-plus-char-shared-secret> \
+  --set selfHosted.bootstrapSubject=<fresh-uuid-v4>
+```
+
+Connect to the AIWorks WebSocket execute endpoint:
+
+```
+ws://<service-or-ingress>/api/repos/<any-uuid>/flows/<flow-slug>/execute
+Authorization: Token <selfHosted.apiKey>
+```
+
+In files mode the `<any-uuid>` is just routing decoration — flows are
+looked up by slug.
+
+## Storage modes
+
+| Mode | Use when | Persistence |
+|---|---|---|
+| `files` (default) | You design flows on cloud (`aiworks.deepgram.com`), export to JSON, drop them in the `flows/` PVC | PVC at `storage.files.flowsDir` |
+| `postgres` | You want the cloud-shape DB-backed flow store | Pass `storage.postgres.dsn` |
+
+See [the self-hosted concepts](https://github.com/deepgram/voiceworks/blob/master/documentation/concepts/self-hosted.md)
+for the two-knobs model and full deployment matrix.
+
+## Examples
+
+### Files mode + open mode (most common)
+
+```sh
+helm install aiworks deepgram/deepgram-aiworks \
+  -f - <<EOF
+license:
+  apiKey: "12345678-1234-5678-abcd-beefdeadbeef"
+  serverUrl:
+    - "http://license-proxy.deepgram.svc.cluster.local:8080"
+selfHosted:
+  apiKey: "a-fresh-shared-secret-of-at-least-forty-chars"
+  bootstrapSubject: "$(uuidgen)"
+storage:
+  mode: files
+  files:
+    persistence:
+      enabled: true
+      size: 5Gi
+deepgram:
+  streamingListenUrl: "ws://api.deepgram.svc.cluster.local:8080/v1/listen"
+  batchListenUrl: "http://api.deepgram.svc.cluster.local:8080/v1/listen"
+EOF
+```
+
+### Behind an ingress
+
+```yaml
+ingress:
+  enabled: true
+  className: nginx
+  annotations:
+    cert-manager.io/cluster-issuer: letsencrypt-prod
+  hosts:
+    - host: aiworks.example.com
+      paths:
+        - path: /
+          pathType: Prefix
+  tls:
+    - hosts: [aiworks.example.com]
+      secretName: aiworks-tls
+```
+
+### In-process TLS (no ingress)
+
+```yaml
+server:
+  tls:
+    enabled: true
+    # Inline (chart creates a kubernetes.io/tls Secret):
+    cert: |
+      -----BEGIN CERTIFICATE-----
+      ...
+    key: |
+      -----BEGIN PRIVATE KEY-----
+      ...
+    # Or reference an existing Secret:
+    # existingSecret: aiworks-tls
+```
+
+SIGHUP triggers cert hot-reload — when the cert file in the mounted
+Secret changes (e.g., cert-manager rotation), send a SIGHUP to the
+pod:
+```sh
+kubectl exec -it aiworks-<id> -- kill -HUP 1
+```
+Or rely on the pod's `terminationGracePeriodSeconds` to drain a
+rolling update.
+
+### Postgres mode
+
+```yaml
+storage:
+  mode: postgres
+  postgres:
+    dsn: "postgres://user:pass@db.example.com:5432/voiceworks"
+```
+
+## Upgrading
+
+```sh
+helm upgrade aiworks deepgram/deepgram-aiworks \
+  --reuse-values \
+  --set image.tag=<new-version>
+```
+
+Default `updateStrategy.rollingUpdate.maxUnavailable=0` keeps at least
+one pod serving during the roll. Long-lived WebSocket sessions are
+drained over `terminationGracePeriodSeconds`.
+
+## Uninstalling
+
+```sh
+helm uninstall aiworks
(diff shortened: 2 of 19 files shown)


mattsloan-dg:
I will be taking a look at this!

charts/deepgram-aiworks/templates/configmap.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+    [server]
+    host   = "{{ .Values.server.host }}"
+    port   = {{ .Values.server.port }}
+    prefix = "{{ .Values.server.prefix }}"
+    ws_idle_timeout = "{{ .Values.server.wsIdleTimeout }}"
+    {{- if .Values.server.tls.enabled }}
+    cert = "/etc/voiceworks/tls/tls.crt"
+    key  = "/etc/voiceworks/tls/tls.key"
+    {{- if or .Values.server.tls.clientCaBundle .Values.server.tls.clientCaBundleExistingSecret }}
```
jkroll-deepgram, line 22: `clientCaBundleExistingSecret` is honored here but nowhere else in the chart — `secret.yaml` only writes `ca.crt` from the inline `clientCaBundle`, and `deployment.yaml` only mounts one TLS volume (the chart-managed or `server.tls.existingSecret`-referenced Secret). If an operator sets `clientCaBundleExistingSecret`, `config.toml` will reference `/etc/voiceworks/tls/ca.crt` but no such file is mounted, and voiceworks will fail to start.

Two ways out:
- Drop `clientCaBundleExistingSecret` and require the CA bundle to live in the same TLS Secret as `tls.crt`/`tls.key` under the key `ca.crt`. Document the expected key names in the README.
- Or add a second volume/volumeMount that pulls `ca.crt` from `clientCaBundleExistingSecret` and projects it next to `tls.crt`.

charts/deepgram-aiworks/values.yaml
```diff
@@ -0,0 +1,200 @@
...
+    failureThreshold: 60   # 5 minutes total
+
+# Termination grace — long enough for in-flight WS sessions to drain.
+terminationGracePeriodSeconds: 60
+
+# Update strategy. maxUnavailable=0 keeps existing pods serving during a roll.
+updateStrategy:
+  type: RollingUpdate
+  rollingUpdate:
+    maxUnavailable: 0
```
jkroll-deepgram, line 191: Heads up: this default interacts badly with the other defaults on this chart. With `replicaCount: 1` (line 17), `storage.files.persistence.enabled: true` (line 46), and `pvc.yaml` hardcoding `accessModes: ["ReadWriteOnce"]`, a rolling update has to schedule the new pod before terminating the old one — but the new pod can't mount the RWO volume while the old one holds it. The roll stalls indefinitely.

Same shape blocks HPA: `scaling.auto.maxReplicas: 5` + files mode + RWO can't scale past 1.

Options:
- Default `maxUnavailable: 1` (correct for `replicaCount: 1`) and let operators tune up when they scale.
- Expose `accessModes` in `values.yaml` so RWX-capable clusters can opt in.
- Add a `fail`/`required` check in templates when the combination is incompatible (`replicas > 1` + files + RWO).

.github/workflows/aiworks-chart-ci.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: helm template
+        run: helm template aiworks-test charts/deepgram-aiworks --values charts/deepgram-aiworks/${{ matrix.values }} > /tmp/rendered.yaml
+
+      - name: Install kubeconform
+        run: |
+          set -e
+          curl -sL https://github.com/yannh/kubeconform/releases/latest/download/kubeconform-linux-amd64.tar.gz | tar -xz -C /usr/local/bin
+
+      - name: kubeconform render
+        run: kubeconform -strict -summary -kubernetes-version 1.28.0 -ignore-missing-schemas /tmp/rendered.yaml
```
jkroll-deepgram, line 40: `-ignore-missing-schemas` silently accepts any kind whose schema isn't in the bundle, including kinds introduced by a typo (e.g., `kind: Deployement`). The only kind in this chart that needs the escape hatch is `ServiceMonitor` (from the prometheus-operator CRDs).

Tighter options:
- `-skip monitoring.coreos.com/v1/ServiceMonitor` — skip just that one kind.
- `-schema-location 'default' -schema-location 'https://raw.githubusercontent.com/datreeio/CRDs-catalog/main/{{.Group}}/{{.ResourceKind}}_{{.ResourceAPIVersion}}.json'` — load community CRD schemas so ServiceMonitor (and others) get validated.

charts/deepgram-aiworks/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,30 @@
...
+spec:
+  podSelector:
+    matchLabels:
+      {{- include "deepgram-aiworks.selectorLabels" . | nindent 6 }}
+  policyTypes:
+    - Ingress
+    {{- if .Values.networkPolicy.egressTo }}
+    - Egress
+    {{- end }}
+  {{- with .Values.networkPolicy.ingressFrom }}
```
jkroll-deepgram, line 17: If an operator sets `networkPolicy.enabled: true` but leaves `ingressFrom: []` (the default in `values.yaml`), the `{{- with ...}}` block elides the entire `ingress:` clause. The rendered Policy then declares `policyTypes: [Ingress]` with no rules — k8s reads that as "deny all ingress to selected pods."

The Service stops accepting any traffic, including from kube-proxy / probes. Failure mode looks like "everything is down" with no obvious cause.

Fixes:
- `{{- required "networkPolicy.ingressFrom must be set when networkPolicy.enabled" .Values.networkPolicy.ingressFrom }}` to fail at template time, or
- Default to an `ingressFrom` rule like `- podSelector: {}` (same namespace) and document the implication, or
- Document loudly in the README that empty `ingressFrom` is default-deny.

charts/deepgram-aiworks/values.yaml
```diff
@@ -0,0 +1,200 @@
...
+
+# License check-in. Required when the image is built with --features self-hosted.
+license:
+  # -- License ID (UUID). Stored in the credentials Secret as DEEPGRAM_API_KEY.
+  # Required unless apiKeyExistingSecret is set.
+  apiKey: ""
+  # -- Alternative: reference an existing Secret rather than letting the chart create one.
+  apiKeyExistingSecret: ""
+  # -- Key inside apiKeyExistingSecret that holds the license UUID.
+  apiKeyExistingSecretKey: "DEEPGRAM_API_KEY"
```
jkroll-deepgram, line 27: This field controls two unrelated things: the key the chart *reads* inside `secrets.existingSecret` (the BYO Secret path) and the key the chart *writes* in its own credentials Secret (`secret.yaml`). If an operator changes it on an existing release with `secrets.existingSecret: ""`, the chart-created Secret renames its key — but `deployment.yaml`'s env ref also follows the rename, so it superficially keeps working until something else (a sidecar, an in-place patch, an external reader) trips over the new key name.

Suggest splitting the responsibility:
- Hardcode the chart-written key names (`DEEPGRAM_API_KEY` and `VW_API_KEY` are standard) inside `secret.yaml` and `deployment.yaml`.
- Keep `*ExistingSecretKey` as a *read-only* knob that only applies when `secrets.existingSecret` is set.

charts/deepgram-aiworks/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,23 @@
...
+    {{- end }}
+spec:
+  selector:
+    matchLabels:
+      {{- include "deepgram-aiworks.selectorLabels" . | nindent 6 }}
+  endpoints:
+    - port: http
+      interval: {{ .Values.serviceMonitor.interval }}
+      scrapeTimeout: {{ .Values.serviceMonitor.scrapeTimeout }}
+      path: /metrics
```
jkroll-deepgram, line 22: This scrapes `/metrics` on the same Service port that handles `/api/health` and the WebSocket execute endpoint. Worth verifying voiceworks actually exposes Prometheus metrics on the main listener — many backends route them onto a separate `:9090` (or `:3001`, etc.) admin port that isn't in the Service or container spec.

If it's a separate port, this ServiceMonitor will scrape voiceworks's app router and either 404 (Prom marks the target down) or, worse, return HTML that breaks the parser. If it's the same port, leaving a comment in `values.yaml` to that effect would save the next operator a debugging session.

charts/deepgram-aiworks/README.md
```diff
@@ -0,0 +1,203 @@
...
+| `selfHosted.apiKey` | string | `""` (required when enabled) | Shared-secret token; stored as `VW_API_KEY` in Secret. **voiceworks resolves `${VW_API_KEY}` from the pod env at startup** — the literal string `"${VW_API_KEY}"` is what gets written to `config.toml`. |
+| `selfHosted.bootstrapSubject` | string | `""` (required when enabled) | Fresh UUID; **never reuse a UUID with existing data** |
+| `deepgram.streamingListenUrl` | string | `ws://api.deepgram.svc.cluster.local:8080/v1/listen` | Deepgram streaming STT endpoint |
+| `deepgram.batchListenUrl` | string | `http://api.deepgram.svc.cluster.local:8080/v1/listen` | Deepgram batch STT endpoint |
+| `server.host` | string | `0.0.0.0` | Bind address |
+| `server.port` | int | `8080` | Bind port |
+| `server.tls.enabled` | bool | `false` | Terminate TLS in voiceworks itself |
+| `server.tls.cert` | string | `""` | PEM cert chain (inline; required when `tls.enabled` and no `existingSecret`) |
+| `server.tls.key` | string | `""` | PEM private key (inline) |
+| `server.tls.existingSecret` | string | `""` | Reference an existing `kubernetes.io/tls` Secret |
```
jkroll-deepgram, line 180: The chart mounts this Secret at `/etc/voiceworks/tls` and references `tls.crt` / `tls.key` / `ca.crt` from `config.toml`. Operators bringing their own Secret need to know which keys to put in it. Suggest tightening this row to something like:

> Reference an existing `kubernetes.io/tls` Secret. Must contain keys `tls.crt`, `tls.key`, and (when `clientCaBundle` is set) `ca.crt`.

Same note belongs on `secrets.existingSecret` (line 197) — that one needs the keys named by `license.apiKeyExistingSecretKey` and `selfHosted.apiKeyExistingSecretKey` (default `DEEPGRAM_API_KEY` and `VW_API_KEY`).

jkroll-deepgram (commented):
Solid bones — sensible security defaults (non-root 10001, read-only rootfs, drop ALL caps), checksum-pinned configmap restart, three meaningful CI scenarios, decent values-reference docs. Inline comments cover the specific bugs / footguns I'd want addressed; this top-level note is for the cross-cutting convention gaps with `deepgram-self-hosted`'s CI.

**Missing repo conventions:**
- **No `CHANGELOG.md`.** The existing chart has one and `ci.yml` fails the PR if a non-doc change to the chart skips the changelog. The new `aiworks-chart-ci.yml` doesn't enforce this. Either add a `CHANGELOG.md` + the equivalent CI check, or accept the inconsistency knowingly.
- **No `helm-docs`.** `ci.yml` regenerates the `deepgram-self-hosted` README from `values.yaml` comments and fails on drift. The new chart's README is hand-written and unenforced — `server.prefix`, `server.wsIdleTimeout`, `storage.postgres.dsnExistingSecret`, `storage.postgres.dsnExistingSecretKey` are already missing from the values-reference table. Recommend adopting `helm-docs` for parity.
- **No chart-version-bump CI check.** Existing CI fails a PR if `appVersion` bumped without `version` bumping. New CI has no equivalent — only `publish.yml` enforces it at release time. Cheap to add and prevents a "merged but didn't publish" surprise.

**Risks worth flagging (already acknowledged in the PR body but restating for the record):**
- Chart depends on voiceworks `feat/self-hosted` landing first for `${VW_API_KEY}` env resolution. Recommend either holding merge until that lands, or tagging this release in a way `publish.yml` won't pick up automatically.
- The "first chart release" branch of `publish.yml` (empty `OLD_VERSION` ⇒ treated as bump) is exercised here for the first time. A dry-run with `act` or `workflow_dispatch` against a throwaway repo would derisk it.

PR #177, opened 2026-07-21
Set LD_LIBRARY_PATH on GPU Engine containers for GKE driver mounts

ehilvanodg:
## Proposed changes

Engine images since `release-260611` (the container refactor) register their bundled libraries via `ldconfig` at build time instead of a baked-in `LD_LIBRARY_PATH`, relying on the NVIDIA container toolkit to inject `libcuda` and refresh the linker cache at container create. GKE with Container-Optimized OS does not use the toolkit: its device plugin mounts the host driver at `/usr/local/nvidia`, and [GCP's GPU documentation](https://cloud.google.com/kubernetes-engine/docs/how-to/gpus#cuda) expects workloads to reference it via `LD_LIBRARY_PATH`. Since the refactor, Engine pods on GKE crash on startup with:

```
impeller: error while loading shared libraries: libcuda.so.1: cannot open shared object file: No such file or directory
```

This was reported in https://github.com/orgs/deepgram/discussions/1626 and has since been hit by multiple GKE self-hosted deployments upgrading from `release-260528` to `release-260611`+ (e.g. chart `0.37.0` → `0.39.0`).

This PR sets `LD_LIBRARY_PATH=/usr/local/nvidia/lib64:/usr/local/nvidia/lib` on GPU-enabled Engine containers — mirroring what pre-refactor images inherited from the CUDA base image, and following the pattern of #166, which moved `NVIDIA_VISIBLE_DEVICES`/`NVIDIA_DRIVER_CAPABILITIES` into the chart after the same refactor:

- The directories do not exist outside GKE, where the dynamic linker silently ignores them → no-op on GPU Operator / container-toolkit clusters.
- The Engine image's own libraries still resolve via its `ldconfig` cache, so there is no interaction with bundled libraries.
- A user-supplied `engine.extraEnv` entry for `LD_LIBRARY_PATH` still takes precedence, since `extraEnv` renders later and Kubernetes uses the last duplicate env entry.
- Not set on CPU-only Engine pods (`requests.gpu: 0`).

Also adds a GKE troubleshooting entry to the chart README (including the `engine.extraEnv` workaround for older chart versions) and a `0.40.2` changelog entry.

## Types of changes

## Checklist

  - `helm lint` passes and `helm-docs` was run per CONTRIBUTING. Verified via a `helm template` render matrix: default GPU Engine pods get the variable; CPU-only pods (`requests.gpu: 0`) do not; `useNvidiaDevicePlugin: true` pods do; a user `engine.extraEnv` override renders after the chart's entry and wins. The equivalent env var (applied via `engine.extraEnv`) is already running in production on GKE deployments affected by this issue; this exact chart build has not yet been re-validated on a live GKE cluster.

## Further comments

Alternatives considered:

- **Image-side fix** (restore a minimal `ENV LD_LIBRARY_PATH=/usr/local/nvidia/lib64:/usr/local/nvidia/lib` in the Engine image, as NVIDIA's own CUDA base images do): more universal, since it also covers non-Helm Kubernetes deployments — but #166 established the chart as the home for environment-compat variables after the container refactor, and the image intentionally moved to `ldconfig`. Happy to coordinate an image-side change instead if preferred; an entrypoint-time `ldconfig` refresh would additionally cover the Flux subprocess on GKE (it launches with `env -u LD_LIBRARY_PATH`, so env-var-based fixes don't reach it and the current workaround is a `postStart` hook).
- **Values-gated flag** (e.g. `engine.gke.enabled`): rejected — the variable is harmless where the paths don't exist, and gating it behind a flag preserves the failure mode for the next GKE deployment.

### DIFF
--- a/charts/deepgram-self-hosted/CHANGELOG.md
+++ b/charts/deepgram-self-hosted/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## Unreleased
 
+## [0.40.2] - 2026-07-21
+
+### Fixed
+
+- Engine pods on GKE with Container-Optimized OS crashed on startup with `libcuda.so.1: cannot open shared object file` when using Engine images `release-260611` or newer. GKE's device plugin mounts the host NVIDIA driver at `/usr/local/nvidia` without the NVIDIA container toolkit, and [expects workloads to reference it via `LD_LIBRARY_PATH`](https://cloud.google.com/kubernetes-engine/docs/how-to/gpus#cuda) — a path older Engine images baked in but newer images do not. The chart now sets `LD_LIBRARY_PATH=/usr/local/nvidia/lib64:/usr/local/nvidia/lib` on GPU-enabled Engine containers; this is a no-op in environments where the NVIDIA container toolkit injects the driver libraries. See [discussion #1626](https://github.com/orgs/deepgram/discussions/1626).
+
 ## [0.40.1] - 2026-07-17
 
 ### Added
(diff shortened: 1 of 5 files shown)


jkroll-deepgram (approved):

jkroll-deepgram:
@therealevanhenry and I confirmed that this change improves usability for GKE customers while not affecting customers on other cloud platforms such as AWS EKS.

PR #182, opened 2026-08-11
Self-hosted release 260812: Flux TTS support & config + tag bump

itayl-dg:
## Proposed changes

- Add Flux TTS config variants (`api.flux-tts.toml`, `engine.flux-tts.toml`) for both standard and license-proxy deploy types, plus Docker and Podman Compose templates. Flux TTS is enabled with a `[flux_tts]` section in `engine.toml` (`enabled`, `uuid`, `max_batch_size`) and `speak_v2` / `speak_v2_streaming` in `api.toml`.
- Add a top-level `fluxTts` block to the Helm chart (`enabled`, `uuid`, `maxBatchSize`) rendering `[flux_tts]` into the Engine ConfigMap, plus `api.features.speakV2` / `speakV2Streaming`. With `agent.enabled`, the config applies only to the text-to-speech Engine. New sample `samples/08-flux-tts-setup.values.yaml`.
- `max_batch_size` defaults to `0`, which is not a valid value. The correct value differs substantially between GPUs, so Engine will not start until it is set explicitly.
- Bump all container tags from `release-260728` to `release-260812`.
- Helm chart 0.41.1 -> 0.42.0 with CHANGELOG entry.

Flux TTS is Early Access in the 260812 self-hosted release and requires `release-260812` or later. Details are in the [main Deepgram changelog](https://developers.deepgram.com/changelog/self-hosted-changelog).

Enabling Flux TTS requires changes on both services. In `engine.toml`, `[flux_tts] enabled = true` requires both `uuid` and `max_batch_size` — Engine refuses to start if either is missing. `max_batch_size` ships as `0`, which is intentionally not a working value: the correct setting differs substantially between GPUs, and one tuned for an L40S will underperform or exhaust memory on an L4. In `api.toml`, `speak_v2` and `speak_v2_streaming` must be enabled, or `/v2/speak` is not exposed even with the Engine configured correctly.

Two further constraints are documented in the chart CHANGELOG and the config comments: Flux TTS and Aura-2 cannot share an Engine (Engine exits at startup if both are configured), and Flux TTS workers bind to a single GPU, so extra GPUs per pod add no capacity.

## Types of changes

## Checklist

  - Not deployed. Verification so far is lint, render, and parse only — no cluster and no EA images.
    - `helm lint` passes; `samples/08-flux-tts-setup.values.yaml` renders the expected `[flux_tts]` section and API feature flags
    - With `agent.enabled`, `[flux_tts]` renders only into the text-to-speech Engine ConfigMap
    - Default render is unchanged, so existing deployments are unaffected
    - `helm-docs` README regenerated
    - No stale `release-260728` tags remain in yml/yaml files


therealevanhenry (approved):
LGTM

PR #183, opened 2026-08-14
Fail render when the Aura-2 GPU request is smaller than cudaVisibleDevices

therealevanhenry:
## Problem

An Aura-2 Engine needs one allocated GPU per index in the active `aura2.*.cudaVisibleDevices`. The chart defaults `engine.resources.requests.gpu` and `.limits.gpu` to `1`, while `aura2.english.cudaVisibleDevices` defaults to `"0,1"` — two devices.

With `engine.resources.useNvidiaDevicePlugin` enabled, the pod sees only the GPUs the device plugin allocates to it, so CUDA device 1 does not exist and Engine panics at startup. The chart rendered that combination without complaint; the failure only surfaced after rollout, as a panic that reads like a model-loading problem.

`engine.agentOverrides` sets `gpu: 2` for the TTS Engine, but only when `agent.enabled: true`. A standalone TTS release gets no such default and has to set the GPU count itself — which is easy to miss, since the values docs described 2 GPUs as a "minimum".

## Change

- **Render-time validation** (`engine.deployment.yaml`): compare the effective per-engine-type GPU request against the number of devices in the active `aura2.*.cudaVisibleDevices` and `fail` when it is short. The message names the values to change, resolving to `engine.agentOverrides.agent-text-to-speech.resources` in voice agent deployments and `engine.resources` otherwise.
- **Values docs**: `Aura-2 requires a minimum of 2 GPUs` → exactly 2, since a larger GPU does not substitute for two and one Engine will not use a third. Call out that `engine.agentOverrides` does not apply outside a voice agent deployment.
- **CHANGELOG + regenerated README** (helm-docs).

### Scope

The check only fires on the device-plugin path, because that is the only configuration that is actually broken. With the plugin disabled (the default) the chart mounts every node GPU into every Engine pod, so a short GPU request still runs — failing it would break deployments that work today. That case keeps rendering and is documented instead, together with the hazard it carries: the pod under-declares GPU usage to the scheduler, which can then place a second Engine on the same node to bind the same physical devices.

## Verification

`helm lint` passes. `helm template` against `samples/04-aura-2-setup.values.yaml` and `samples/05-voice-agent-aws.values.yaml`:

| Case | Result |
| --- | --- |
| plugin on, `gpu: 2`, devices `"0,1"` | renders |
| plugin on, `gpu: 1`, devices `"0,1"` | fails, names `engine.resources.*` |
| plugin **off**, `gpu: 1`, devices `"0,1"` | renders (no regression) |
| defaults, `aura2` disabled | renders |
| agent mode, plugin on, override `gpu: 2` | renders |
| agent mode, plugin on, override `gpu: 1` | fails, names `engine.agentOverrides.agent-text-to-speech.resources.*` |

🤖 Generated with [Claude Code](https://claude.com/claude-code)


charts/deepgram-self-hosted/values.yaml
```diff
@@ -556,12 +556,20 @@ engine:
...
+    # place a second Engine on the same node to bind the same physical devices.
+    # See: https://github.com/NVIDIA/k8s-device-plugin
     useNvidiaDevicePlugin: false
     requests:
       memory: "30Gi"
       cpu: "4000m"
-      # -- Number of GPUs to request
+      # -- Number of GPUs to request. Aura and Aura-2 need 2 GPUs per Engine, and
+      # `engine.agentOverrides` does not apply outside a voice agent deployment, so a
+      # standalone TTS release must set this (and the matching limit) to 2 explicitly.
```
pcgeek86, line 572: The comment here says that this must be set to 2 explicitly, but it's set to 1. Is that okay?

charts/deepgram-self-hosted/templates/engine/engine.deployment.yaml
```diff
@@ -179,6 +179,21 @@ spec:
...
+        {{- else if $.Values.aura2.polyglot.enabled }}
+        {{- $aura2Block = $.Values.aura2.polyglot }}
+        {{- end }}
+        {{- if and $.Values.engine.resources.useNvidiaDevicePlugin $aura2Block.cudaVisibleDevices }}
+        {{- $deviceCount := len (splitList "," (trim $aura2Block.cudaVisibleDevices)) }}
+        {{- if lt $gpuRequest $deviceCount }}
+        {{- $gpuKey := ternary (printf "engine.agentOverrides.%s.resources" $type) "engine.resources" (ne $type "") }}
+        {{- fail (printf "Error: Aura-2 needs one allocated GPU per index in cudaVisibleDevices (%q is %d devices), but this Engine requests %d. engine.resources.useNvidiaDevicePlugin is enabled, so the pod only sees the GPUs the device plugin allocates to it, and Engine panics at startup on a device that does not exist. Set %s.requests.gpu and %s.limits.gpu to %d, or reduce cudaVisibleDevices to the devices you allocate." $aura2Block.cudaVisibleDevices $deviceCount $gpuRequest $gpuKey $gpuKey $deviceCount) }}
+        {{- end }}
+        {{- end }}
```
jkroll-deepgram: Index-based rather than count-based, plus the checks that hold in every mode: the standalone `gpu: 1` default, and a request/limit mismatch (Kubernetes requires extended-resource request and limit to be equal, so that currently passes render and fails at apply).

Floor of 2 rather than exactly 2, since requesting 3+ wastes a device but works, and claiming a whole 4-GPU node for one Engine is a legitimate anti-co-tenancy pattern. This also drops "or reduce cudaVisibleDevices to the devices you allocate" — with a 2-GPU requirement that trades a panic for a deployment that can't serve.

```suggestion
        {{- $aura2Block := dict }}
        {{- if $.Values.aura2.english.enabled }}
        {{- $aura2Block = $.Values.aura2.english }}
        {{- else if $.Values.aura2.spanish.enabled }}
        {{- $aura2Block = $.Values.aura2.spanish }}
        {{- else if $.Values.aura2.polyglot.enabled }}
        {{- $aura2Block = $.Values.aura2.polyglot }}
        {{- end }}
        {{- $gpuKey := ternary (printf "engine.agentOverrides.%s.resources" $type) "engine.resources" (ne $type "") }}
        {{- if not $aura2Block }}
        {{- fail "Error: aura2.enabled is true but no language block is enabled, so no Aura-2 configuration is rendered. Enable one of aura2.english, aura2.spanish, or aura2.polyglot." }}
        {{- end }}
        {{- if lt $gpuRequest 2 }}
        {{- fail (printf "Error: Aura-2 needs 2 dedicated GPUs per Engine, but this Engine requests %d. Set %s.requests.gpu and %s.limits.gpu to 2. A single larger GPU does not substitute for two." $gpuRequest $gpuKey $gpuKey) }}
        {{- end }}
        {{- if ne $gpuRequest $gpuLimit }}
        {{- fail (printf "Error: %s.requests.gpu (%d) and %s.limits.gpu (%d) must be equal. Kubernetes rejects pods whose extended-resource request and limit differ." $gpuKey $gpuRequest $gpuKey $gpuLimit) }}
        {{- end }}
        {{- if and $.Values.engine.resources.useNvidiaDevicePlugin $aura2Block.cudaVisibleDevices }}
        {{- range $device := splitList "," $aura2Block.cudaVisibleDevices }}
        {{- $index := trim $device }}
        {{- if not (regexMatch "^[0-9]+$" $index) }}
        {{- fail (printf "Error: cudaVisibleDevices must be a comma-separated list of non-negative integers, got %q." $aura2Block.cudaVisibleDevices) }}
        {{- end }}
        {{- if ge (atoi $index) $gpuRequest }}
        {{- fail (printf "Error: cudaVisibleDevices %q refers to device %s, but engine.resources.useNvidiaDevicePlugin re-indexes each Engine's allocated GPUs from 0, so this pod only sees devices 0-%d and Engine panics at startup on a device that does not exist. Set cudaVisibleDevices to \"0,1\" and let the device plugin choose the physical GPUs." $aura2Block.cudaVisibleDevices $index (sub $gpuRequest 1)) }}
        {{- end }}
        {{- end }}
        {{- end }}
```

charts/deepgram-self-hosted/values.yaml
```diff
@@ -556,12 +556,20 @@ engine:
     # Note: with this enabled, each pod sees only its allocated GPUs (re-indexed
     # from 0), so any `aura2.*.cudaVisibleDevices` values are interpreted relative
     # to that set. Leave them at defaults and size the GPU request to match
-    # (specifically, Aura-2 needs 2 GPUs). See: https://github.com/NVIDIA/k8s-device-plugin
+    # (specifically, Aura-2 needs 2 GPUs); the chart fails to render if the request
+    # is smaller than the number of devices listed in `aura2.*.cudaVisibleDevices`.
+    # When disabled (the default), the chart mounts all of the node's GPUs into
+    # every Engine pod. A GPU request smaller than the device count still runs in
+    # that mode, but it under-declares GPU usage to the scheduler, which can then
+    # place a second Engine on the same node to bind the same physical devices.
```
jkroll-deepgram: "Leave them at defaults" is the one thing not to do here: `aura2.spanish` and `aura2.polyglot` default to `"2,3"`, which is never valid under the device plugin.

```suggestion
    # Note: with this enabled, each pod sees only its allocated GPUs (re-indexed
    # from 0), so `aura2.*.cudaVisibleDevices` must list indices below the GPU
    # request — use `"0,1"` and let the device plugin choose the physical GPUs.
    # The `"2,3"` defaults on `aura2.spanish`/`aura2.polyglot` are for the
    # plugin-off case below and are never valid with the plugin enabled; the
    # chart fails to render on an out-of-range index.
    # When disabled (the default), the chart mounts all of the node's GPUs into
    # every Engine pod, so `"2,3"` silently requires a 4-GPU node. A GPU request
    # smaller than the device count still runs in that mode, but it under-declares
    # GPU usage to the scheduler, which can then place a second Engine on the same
    # node to bind the same physical devices.
```

charts/deepgram-self-hosted/values.yaml
jkroll-deepgram: Aura-1 needs 1 GPU, not 2, and the chart has no Aura-1 block, so the claim covers values that don't exist here. This key is also the global default for every Engine type, so an STT-only reader comes away thinking they need 2, and a standalone Flux TTS release should stay at 1.

```suggestion
      # -- Number of GPUs to request. Applies to all Engine types. Aura-2 needs 2
      # GPUs per Engine; Flux TTS needs 1 dedicated GPU. `engine.agentOverrides`
      # applies only when `agent.enabled` is true, so a standalone Aura-2 release
      # must set this (and the matching limit) to 2 explicitly.
```

charts/deepgram-self-hosted/CHANGELOG.md
```diff
@@ -6,6 +6,18 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## Unreleased
 
+### Added
+
+- Added render-time validation for the Aura-2 GPU count. With `engine.resources.useNvidiaDevicePlugin` enabled, an Engine pod sees only the GPUs the device plugin allocates to it, so a GPU request smaller than the number of devices listed in the active `aura2.*.cudaVisibleDevices` leaves Engine panicking at startup on a device that does not exist. That combination now fails `helm install`/`helm upgrade` with a message naming the values to change, resolving to the `engine.agentOverrides.agent-text-to-speech.resources` path in voice agent deployments and to `engine.resources` otherwise.
+
+### Changed
+
+- Documented that Aura and Aura-2 need exactly 2 GPUs per Engine: a single larger GPU does not substitute for two, and one Engine will not schedule work across more than two. Additional GPUs on a machine serve additional Engine instances. Because `engine.agentOverrides` applies only when `agent.enabled` is true, a standalone TTS release must set `engine.resources.requests.gpu` and `engine.resources.limits.gpu` to `2` itself.
```
jkroll-deepgram: Aura-1 needs 1 GPU. Also scoping "exactly" to a floor, since 3+ wastes a device but doesn't break.

```suggestion
- Documented that an Aura-2 Engine needs 2 dedicated GPUs: a single larger GPU does not substitute for two, and one Engine will not schedule work across more than two, so additional GPUs on a machine serve additional Engine instances. Flux TTS needs 1 dedicated GPU. Because `engine.agentOverrides` applies only when `agent.enabled` is true, a standalone Aura-2 release must set `engine.resources.requests.gpu` and `engine.resources.limits.gpu` to `2` itself.
```

charts/deepgram-self-hosted/templates/engine/engine.deployment.yaml
jkroll-deepgram: Compares index values rather than device count, and adds the checks that apply regardless of the device plugin: a GPU request below 2, a device list with fewer than two entries, and a request/limit mismatch (Kubernetes requires extended-resource requests and limits to be equal, so a mismatch currently renders and is then rejected at apply time).

A floor of 2 rather than an equality check, because a request above 2 leaves a GPU unused but works, and requesting all of a node's GPUs for one Engine to avoid co-tenancy is a valid configuration.

`toString` and `default` on the device list are needed because an unquoted `cudaVisibleDevices: 0` and an empty string are both falsy, so a truthiness guard skips the device checks while the template still emits `CUDA_VISIBLE_DEVICES`.

This also removes "or reduce cudaVisibleDevices to the devices you allocate" from the message, since with a 2-GPU requirement that produces a deployment that cannot serve.

```suggestion
        {{- $aura2Block := dict }}
        {{- if $.Values.aura2.english.enabled }}
        {{- $aura2Block = $.Values.aura2.english }}
        {{- else if $.Values.aura2.spanish.enabled }}
        {{- $aura2Block = $.Values.aura2.spanish }}
        {{- else if $.Values.aura2.polyglot.enabled }}
        {{- $aura2Block = $.Values.aura2.polyglot }}
        {{- end }}
        {{- $gpuKey := ternary (printf "engine.agentOverrides.%s.resources" $type) "engine.resources" (ne $type "") }}
        {{- if not $aura2Block }}
        {{- fail "Error: aura2.enabled is true but no language block is enabled, so no Aura-2 configuration is rendered. Enable one of aura2.english, aura2.spanish, or aura2.polyglot." }}
        {{- end }}
        {{- if lt $gpuRequest 2 }}
        {{- fail (printf "Error: Aura-2 needs 2 dedicated GPUs per Engine, but this Engine requests %d. Set %s.requests.gpu and %s.limits.gpu to 2. A single larger GPU does not substitute for two." $gpuRequest $gpuKey $gpuKey) }}
        {{- end }}
        {{- if ne $gpuRequest $gpuLimit }}
        {{- fail (printf "Error: %s.requests.gpu (%d) and %s.limits.gpu (%d) must be equal. Kubernetes rejects pods whose extended-resource request and limit differ." $gpuKey $gpuRequest $gpuKey $gpuLimit) }}
        {{- end }}
        {{- $rawDevices := toString (default "" $aura2Block.cudaVisibleDevices) }}
        {{- $devices := splitList "," $rawDevices }}
        {{- if lt (len $devices) 2 }}
        {{- fail (printf "Error: Aura-2 needs 2 dedicated GPUs, but cudaVisibleDevices is %q. Set it to two quoted device indices, e.g. \"0,1\"." $rawDevices) }}
        {{- end }}
        {{- range $device := $devices }}
        {{- $index := trim $device }}
        {{- if not (regexMatch "^[0-9]+$" $index) }}
        {{- fail (printf "Error: cudaVisibleDevices must be a comma-separated list of non-negative integers, got %q." $rawDevices) }}
        {{- end }}
        {{- if and $.Values.engine.resources.useNvidiaDevicePlugin (ge (atoi $index) $gpuRequest) }}
        {{- fail (printf "Error: cudaVisibleDevices %q refers to device %s, but engine.resources.useNvidiaDevicePlugin re-indexes each Engine's allocated GPUs from 0, so this pod only sees devices 0-%d and Engine panics at startup on a device that does not exist. Set cudaVisibleDevices to \"0,1\" and let the device plugin choose the physical GPUs." $rawDevices $index (sub $gpuRequest 1)) }}
        {{- end }}
        {{- end }}
```

charts/deepgram-self-hosted/values.yaml
jkroll-deepgram: `aura2.spanish` and `aura2.polyglot` default to `"2,3"`, which is out of range under the device plugin, so "Leave them at defaults" does not hold for those two blocks.

```suggestion
    # Note: with this enabled, each pod sees only its allocated GPUs (re-indexed
    # from 0), so `aura2.*.cudaVisibleDevices` must list indices below the GPU
    # request — use `"0,1"` and let the device plugin choose the physical GPUs.
    # The `"2,3"` defaults on `aura2.spanish`/`aura2.polyglot` are for the
    # plugin-off case below and are never valid with the plugin enabled; the
    # chart fails to render on an out-of-range index.
    # When disabled (the default), the chart mounts all of the node's GPUs into
    # every Engine pod, so `"2,3"` silently requires a 4-GPU node. A GPU request
    # smaller than the device count still runs in that mode, but it under-declares
    # GPU usage to the scheduler, which can then place a second Engine on the same
    # node to bind the same physical devices.
```

charts/deepgram-self-hosted/values.yaml
jkroll-deepgram: Aura-1 requires 1 GPU. The chart has no Aura-1 configuration, so that part of the sentence refers to values that do not exist here. This key is also the global default for all Engine types, so it reads as a 2-GPU requirement for STT-only deployments, and a standalone Flux TTS release should leave it at 1.

```suggestion
      # -- Number of GPUs to request. Applies to all Engine types. Aura-2 needs 2
      # GPUs per Engine; Flux TTS needs 1 dedicated GPU. `engine.agentOverrides`
      # applies only when `agent.enabled` is true, so a standalone Aura-2 release
      # must set this (and the matching limit) to 2 explicitly.
```

charts/deepgram-self-hosted/values.yaml
```diff
@@ -574,8 +582,10 @@ engine:
   # Supported keys: agent-speech-to-text, agent-text-to-speech, agent-end-of-turn
   #
   # The agent-text-to-speech engine defaults to gpu: 2 because Aura-2 TTS requires
-  # a minimum of 2 GPUs to serve traffic. STT and EOT engines are unaffected and
-  # continue to use the global engine.resources values (gpu: 1 by default).
+  # exactly 2 GPUs to serve traffic: a single larger GPU does not substitute for two,
+  # and one Engine will not schedule work across more than two. STT and EOT engines
+  # are unaffected and continue to use the global engine.resources values (gpu: 1 by
+  # default).
```
jkroll-deepgram: Same "exactly 2" wording as the CHANGELOG entry. Suggesting a floor here as well so the three locations agree.

```suggestion
  # The agent-text-to-speech engine defaults to gpu: 2 because Aura-2 TTS needs
  # 2 dedicated GPUs to serve traffic: a single larger GPU does not substitute for
  # two, and one Engine will not schedule work across more than two, so anything
  # above 2 goes unused. STT and EOT engines are unaffected and continue to use
  # the global engine.resources values (gpu: 1 by default).
```

charts/deepgram-self-hosted/CHANGELOG.md
```diff
@@ -6,6 +6,18 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## Unreleased
 
+### Added
+
+- Added render-time validation for the Aura-2 GPU count. With `engine.resources.useNvidiaDevicePlugin` enabled, an Engine pod sees only the GPUs the device plugin allocates to it, so a GPU request smaller than the number of devices listed in the active `aura2.*.cudaVisibleDevices` leaves Engine panicking at startup on a device that does not exist. That combination now fails `helm install`/`helm upgrade` with a message naming the values to change, resolving to the `engine.agentOverrides.agent-text-to-speech.resources` path in voice agent deployments and to `engine.resources` otherwise.
```
jkroll-deepgram: Describes the count-based, device-plugin-only check. Updated to match the behaviour in the suggestion above.

```suggestion
- Added render-time validation for the Aura-2 GPU configuration. An Aura-2 Engine needs 2 dedicated GPUs, so `helm install`/`helm upgrade` now fails when the effective GPU request for an Aura-2 Engine is below 2, when its request and limit differ, when the active `aura2.*.cudaVisibleDevices` lists fewer than two device indices or a non-integer value, or when `aura2.enabled` is true with no language block enabled. With `engine.resources.useNvidiaDevicePlugin` enabled an Engine pod sees only the GPUs the device plugin allocates to it, re-indexed from 0, so the chart additionally fails when any listed index falls outside that range — previously that combination rendered without complaint and left Engine panicking at startup on a device that does not exist. Messages name the values to change, resolving to `engine.agentOverrides.agent-text-to-speech.resources` in voice agent deployments and to `engine.resources` otherwise.
```

charts/deepgram-self-hosted/CHANGELOG.md
jkroll-deepgram: Aura-1 requires 1 GPU, and "exactly" overstates the upper bound. This also adds a breaking-change entry, since the request floor applies with the device plugin disabled; the wording follows the format already used at line 109.

```suggestion
- Documented that an Aura-2 Engine needs 2 dedicated GPUs: a single larger GPU does not substitute for two, and one Engine will not schedule work across more than two, so additional GPUs on a machine serve additional Engine instances rather than one larger Engine. Flux TTS needs 1 dedicated GPU. Because `engine.agentOverrides` applies only when `agent.enabled` is true, a standalone Aura-2 release must set `engine.resources.requests.gpu` and `engine.resources.limits.gpu` to `2` itself.
- **Breaking change for standalone Aura-2 deployments:** A release left at the default `engine.resources.requests.gpu: 1` now fails to render instead of deploying an Engine that under-declares its GPU usage. Set `engine.resources.requests.gpu` and `engine.resources.limits.gpu` to `2`.
```

charts/deepgram-self-hosted/CHANGELOG.md
```diff
@@ -6,6 +6,18 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
...
+
+### Changed
+
+- Documented that Aura and Aura-2 need exactly 2 GPUs per Engine: a single larger GPU does not substitute for two, and one Engine will not schedule work across more than two. Additional GPUs on a machine serve additional Engine instances. Because `engine.agentOverrides` applies only when `agent.enabled` is true, a standalone TTS release must set `engine.resources.requests.gpu` and `engine.resources.limits.gpu` to `2` itself.
+
+### Notes
+
+- With `engine.resources.useNvidiaDevicePlugin` disabled (the default), the chart mounts every node GPU into every Engine pod, so a GPU request smaller than the `cudaVisibleDevices` device count still runs and is left as-is by the new validation. It does under-declare GPU usage to the scheduler, which can then place a second Engine on the same node and bind the same physical devices. Size the request to the devices the Engine actually uses, or enable the device plugin.
```
jkroll-deepgram: Under the suggested check this is no longer accurate: the request floor of 2 applies with the device plugin disabled as well, so a short request is not left as-is. Only the index-range check is scoped to the device plugin.

```suggestion
- With `engine.resources.useNvidiaDevicePlugin` disabled (the default), the chart mounts every node GPU into every Engine pod, so whether a `cudaVisibleDevices` index resolves depends on the node's physical GPU count, which the chart cannot see at render time; only the device-plugin path is validated for index range. The `"2,3"` defaults on `aura2.spanish`/`aura2.polyglot` therefore require a 4-GPU node in this mode. Sizing the GPU request below the devices an Engine actually uses also under-declares GPU usage to the scheduler, which can then place a second Engine on the same node to bind the same physical devices.
```

charts/deepgram-self-hosted/CHANGELOG.md
jkroll-deepgram: Describes the count-based, device-plugin-only check. Updated to match the behavior in the suggestion above.

```suggestion
- Added render-time validation for the Aura-2 GPU configuration. An Aura-2 Engine needs 2 dedicated GPUs, so `helm install`/`helm upgrade` now fails when the effective GPU request for an Aura-2 Engine is below 2, when its request and limit differ, when the active `aura2.*.cudaVisibleDevices` lists fewer than two device indices or a non-integer value, or when `aura2.enabled` is true with no language block enabled. With `engine.resources.useNvidiaDevicePlugin` enabled an Engine pod sees only the GPUs the device plugin allocates to it, re-indexed from 0, so the chart additionally fails when any listed index falls outside that range — previously that combination rendered without complaint and left Engine panicking at startup on a device that does not exist. Messages name the values to change, resolving to `engine.agentOverrides.agent-text-to-speech.resources` in voice agent deployments and to `engine.resources` otherwise.
```

charts/deepgram-self-hosted/CHANGELOG.md
jkroll-deepgram: Aura-1 requires 1 GPU, and "exactly" overstates the upper bound. This also adds a breaking-change entry, since the request floor applies with the device plugin disabled; the wording follows the format already used at line 109.

```suggestion
- Documented that an Aura-2 Engine needs 2 dedicated GPUs: a single larger GPU does not substitute for two, and one Engine will not schedule work across more than two, so additional GPUs on a machine serve additional Engine instances rather than one larger Engine. Flux TTS needs 1 dedicated GPU. Because `engine.agentOverrides` applies only when `agent.enabled` is true, a standalone Aura-2 release must set `engine.resources.requests.gpu` and `engine.resources.limits.gpu` to `2` itself.
- **Breaking change for standalone Aura-2 deployments:** A release left at the default `engine.resources.requests.gpu: 1` now fails to render instead of deploying an Engine that under-declares its GPU usage. Set `engine.resources.requests.gpu` and `engine.resources.limits.gpu` to `2`.
```

charts/deepgram-self-hosted/CHANGELOG.md
```diff
@@ -6,6 +6,18 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## Unreleased
 
+### Added
```
jkroll-deepgram: The last four PRs (#175, #177, #180, #182) each added a dated `## [x.y.z]` section and bumped `version` in `Chart.yaml`; `## Unreleased` has been an empty placeholder on main throughout. This PR adds its entries under `## Unreleased` and leaves `Chart.yaml` at `0.42.0`, so as it stands the change can't be published as a release.

`Chart.yaml` isn't in the diff so there's no line to attach to, but it needs `version: 0.42.1` alongside this. Adjust the date to the merge date.

```suggestion
## [0.42.1] - 2026-08-19

### Added
```

charts/deepgram-self-hosted/templates/engine/engine.deployment.yaml
jkroll-deepgram: Same structure and gating, with the highest device index in place of the device count. For `"2,3"` this asks for 4 allocated GPUs rather than 2, which is what the device plugin actually has to hand the pod for those indices to exist.

This also replaces "reduce cudaVisibleDevices to the devices you allocate" with setting it to `"0,1"`. Reducing it to one device would satisfy the old check but leave an Aura-2 Engine with a single GPU.

Line 11 of the CHANGELOG describes this as a comparison against the number of devices listed, so it needs a corresponding word change.

```suggestion
        {{- if and $.Values.engine.resources.useNvidiaDevicePlugin $aura2Block.cudaVisibleDevices }}
        {{- $gpusNeeded := 0 }}
        {{- range $device := splitList "," $aura2Block.cudaVisibleDevices }}
        {{- $gpusNeeded = max $gpusNeeded (add1 (atoi (trim $device))) }}
        {{- end }}
        {{- if lt $gpuRequest $gpusNeeded }}
        {{- $gpuKey := ternary (printf "engine.agentOverrides.%s.resources" $type) "engine.resources" (ne $type "") }}
        {{- fail (printf "Error: engine.resources.useNvidiaDevicePlugin is enabled, so an Engine pod only sees the GPUs the device plugin allocates to it, re-indexed from 0. cudaVisibleDevices %q needs %d allocated GPUs, but this Engine requests %d, so Engine panics at startup on a device that does not exist. Either set %s.requests.gpu and %s.limits.gpu to %d, or set cudaVisibleDevices to \"0,1\" and let the device plugin choose which physical GPUs to attach." $aura2Block.cudaVisibleDevices $gpusNeeded $gpuRequest $gpuKey $gpuKey $gpusNeeded) }}
        {{- end }}
        {{- end }}
```

charts/deepgram-self-hosted/values.yaml
jkroll-deepgram: Aura-1 requires 1 GPU, and the chart has no Aura-1 configuration, so that clause refers to values that don't exist here. This key is also the global default for every Engine type, so it reads as a 2-GPU requirement for STT-only deployments, and a standalone Flux TTS release should stay at 1.

```suggestion
      # -- Number of GPUs to request. Applies to all Engine types. Aura-2 needs 2
      # GPUs per Engine; Flux TTS needs 1 dedicated GPU. `engine.agentOverrides`
      # applies only when `agent.enabled` is true, so a standalone Aura-2 release
      # must set this (and the matching limit) to 2 explicitly.
```

charts/deepgram-self-hosted/CHANGELOG.md
jkroll-deepgram: Same Aura-1 point as in `values.yaml`. Adding the Flux TTS count too, since #182 just landed it.

```suggestion
- Documented that an Aura-2 Engine needs exactly 2 GPUs: a single larger GPU does not substitute for two, and one Engine will not schedule work across more than two. Additional GPUs on a machine serve additional Engine instances. Flux TTS needs 1 dedicated GPU. Because `engine.agentOverrides` applies only when `agent.enabled` is true, a standalone Aura-2 release must set `engine.resources.requests.gpu` and `engine.resources.limits.gpu` to `2` itself.
```

charts/deepgram-self-hosted/templates/engine/engine.deployment.yaml
jkroll-deepgram: Three checks, in order of how much the Engine source supports them.

The device-list length and the integer check don't depend on the device plugin, because the assert is on the number of devices CUDA resolves, not on the GPU request. `cudaVisibleDevices: "0"` resolves one device on a 4-GPU node and aborts regardless of what the pod requested, and CUDA ignores an unparseable entry along with every index after it. Neither of these breaks a configuration that works today: any value with fewer than two resolvable indices is already aborting.

The index-vs-request check stays gated on the device plugin, since that's the only mode where the chart knows how many devices the pod will see. It replaces the device count with the highest index, so `"2,3"` asks for 4 allocated GPUs rather than 2.

The messages name the abort Engine actually produces so it can be searched. The original "reduce cudaVisibleDevices to the devices you allocate" would satisfy the old check with a single device, which Aura-2 can't run on.

Line 11 of the CHANGELOG describes this as a comparison against the number of devices listed, so it needs a corresponding word change.

```suggestion
        {{- $gpuKey := ternary (printf "engine.agentOverrides.%s.resources" $type) "engine.resources" (ne $type "") }}
        {{- $devices := splitList "," (toString (default "" $aura2Block.cudaVisibleDevices)) }}
        {{- if lt (len $devices) 2 }}
        {{- fail (printf "Error: Aura-2 needs 2 visible CUDA devices, but cudaVisibleDevices is %q. Engine keeps Code2Code and Code2Audio on one dedicated device and Prompt2Code on another, and aborts at startup with \"insufficient devices for inferencer configuration\" when fewer are visible. Set it to two device indices, e.g. \"0,1\"." (toString (default "" $aura2Block.cudaVisibleDevices))) }}
        {{- end }}
        {{- $gpusNeeded := 0 }}
        {{- range $device := $devices }}
        {{- if not (regexMatch "^[0-9]+$" (trim $device)) }}
        {{- fail (printf "Error: cudaVisibleDevices must be a comma-separated list of non-negative device indices, got %q. CUDA ignores an unparseable entry and every index after it, leaving Engine with fewer visible devices than Aura-2 needs." $aura2Block.cudaVisibleDevices) }}
        {{- end }}
        {{- $gpusNeeded = max $gpusNeeded (add1 (atoi (trim $device))) }}
        {{- end }}
        {{- if and $.Values.engine.resources.useNvidiaDevicePlugin (lt $gpuRequest $gpusNeeded) }}
        {{- fail (printf "Error: engine.resources.useNvidiaDevicePlugin is enabled, so an Engine pod only sees the GPUs the device plugin allocates to it, re-indexed from 0. cudaVisibleDevices %q needs %d allocated GPUs but this Engine requests %d, so Engine aborts at startup with \"insufficient devices for inferencer configuration\". Aura-2 needs 2 visible devices: set cudaVisibleDevices to \"0,1\" and %s.requests.gpu and %s.limits.gpu to 2." $aura2Block.cudaVisibleDevices $gpusNeeded $gpuRequest $gpuKey $gpuKey) }}
        {{- end }}
```

charts/deepgram-self-hosted/values.yaml
jkroll-deepgram: Aura-1 requires 1 GPU, and the chart has no Aura-1 configuration, so that clause refers to values that don't exist here. This key is also the global default for every Engine type, so it reads as a 2-GPU requirement for STT-only deployments, and a standalone Flux TTS release should stay at 1.

```suggestion
      # -- Number of GPUs to request. Applies to all Engine types. Aura-2 needs 2
      # GPUs per Engine; Flux TTS needs 1 dedicated GPU. `engine.agentOverrides`
      # applies only when `agent.enabled` is true, so a standalone Aura-2 release
      # must set this (and the matching limit) to 2 explicitly.
```

charts/deepgram-self-hosted/CHANGELOG.md
jkroll-deepgram: Same Aura-1 point as in `values.yaml`. Adding the Flux TTS count too, since #182 just landed it.

```suggestion
- Documented that an Aura-2 Engine needs exactly 2 GPUs: a single larger GPU does not substitute for two, and one Engine will not schedule work across more than two. Additional GPUs on a machine serve additional Engine instances. Flux TTS needs 1 dedicated GPU. Because `engine.agentOverrides` applies only when `agent.enabled` is true, a standalone Aura-2 release must set `engine.resources.requests.gpu` and `engine.resources.limits.gpu` to `2` itself.
```

PR #184, opened 2026-08-18
Add global.fips.enabled and fix readiness probes on -fips images

jkroll-deepgram:
## Summary

Two changes for FIPS 140-3 deployments, both affecting the `-fips` image variants
introduced in `release-260728`:

1. `grep --quiet` → `grep -q` in the API and License Proxy readiness probes.
   BusyBox grep supports no GNU long options and exits 2 with a usage dump on
   `--quiet`. The `-fips` images have `/usr/bin/grep` symlinked to `/bin/busybox`,
   so both probes always failed and the pods never became ready.

2. New `global.fips.enabled` value (default `false`) that renders
   `[fips] mode = "enabled"` into `api.toml`, `engine.toml`,
   `license-proxy.toml`, and `billing.toml`.

The chart previously rendered no `[fips]` section for any component. Since the
FIPS images do not enable FIPS mode on their own, anyone deploying them via this
chart was silently running OpenSSL in standard mode — the failure is quiet, and
these are deployments chosen specifically to satisfy an audit requirement.

Chart version `0.42.0` → `0.43.0`.

## Verification

Run against the real `quay.io/deepgram/self-hosted-license-proxy:release-260812-fips`
and `self-hosted-api:release-260812-fips` images:

| | `--quiet` | `-q` |
|---|---|---|
| BusyBox (`-fips`, v1.37.0) | exit 2 + usage dump | exit 0 |
| GNU grep 3.12 (non-FIPS, Rocky 9.8) | exit 0 | exit 0 |

- Feeding the chart-rendered `license-proxy.toml` to the FIPS image logs
  `openssl_fips_enabled=true` with the flag on and `false` with it off.
- Every rendered TOML parses with `tomllib`, FIPS on and off, including the
  Billing and Voice Agent (3 Engine ConfigMap) paths.
- With `global.fips.enabled=false` and the version held constant, the rendered
  output is byte-identical to `main` apart from the two `grep` lines — no config
  checksum churn, so upgrades don't needlessly roll pods.
- `helm lint` clean; the existing `[flux_tts]` guard still fires.

## Notes for reviewers

- `[fips]` placement in `engine.toml` is deliberate. It goes after `[license]`
  because `max_active_requests` is a bare top-level key — a section header above
  it would silently absorb it into the FIPS table, producing valid TOML that
  means the wrong thing.
- The flag does not change image tags. Enabling it without `-fips` images gives
  FIPS mode on a non-FIPS image. Tags stay user-controlled because mirrored
  registries use their own naming; the requirement is documented in `values.yaml`.
- Single global toggle rather than per-component, since FIPS is a
  deployment-wide compliance posture and per-component flags invite partial
  compliance.
- `[fips]` in `api.customToml` / `engine.customToml` now fails at render time
  when the flag is set, mirroring the `[flux_tts]` guard.
- Also fixes two pre-existing CHANGELOG issues: `[0.42.0]` had no compare link,
  and `[unreleased]` still compared from `0.41.1`.


jcdyer (approved):
This looks reasonable to me.  I'm not particularly versed in helm charts, but if this passes manual testing on all four applications, there's not much else that would worry me.

therealevanhenry (changes requested):
Rendering fails when `global.fips` is absent from values:

```
$ helm template t ./deepgram-self-hosted --set global.fips=null
engine.config.yaml:58:12: nil pointer evaluating interface {}.enabled
```

Same failure from a values file containing `global: {fips:}`. All four config templates and both `customToml` guards hit it. `main` renders fine under the same inputs.

`helm lint` passes on this branch, and `ci.yml` runs no `helm template`, so CI doesn't cover it. `publish.yml` releases on push to `main` when the chart version is bumped.

`($.Values.global.fips).enabled` in all six spots renders byte-identical output with the key present, flag on and off, and renders when it's absent. `api.config.yaml:67` and `engine.config.yaml:39` already use `default` and `default dict` for the same shape; the comment on the former names `helm upgrade --reuse-values` as the case.

Questions:

1. `values.yaml` says enabling the flag without `-fips` images "gives FIPS mode on a non-FIPS image." aegis `install_crypto_provider` calls `rustls_openssl::fips::enable()` under `if fips_mode.enabled()`, and its doc comment states it panics when libssl isn't a FIPS build. Have you run flag-on against a non-FIPS image to see which happens?
2. In the BusyBox `-q` row, did the input match? The license-proxy pattern uses `\(Connected\|TrustBased\)`, and the table reports the flag's exit code.
3. `grep -rn fips` returns nothing outside `charts/`, including `common/standard_deploy/*.toml` and `common/license_proxy_deploy/*.toml`. Intentional scope for this PR?

Two things I checked against the open questions from the thread. `install_crypto_provider(config.fips.mode)` appears in `src/lib.rs` on `origin/main` for stem, impeller, hermes, and athena. And with `concurrencyLimit.activeRequests` set and the flag on, `tomllib` parses `max_active_requests` as a bare top-level key, not under `fips`.

therealevanhenry (approved):

jkroll-deepgram:
Thanks @therealevanhenry — all three answered.

Non-FIPS images: I had not tested this, and you were right to ask. Ran it natively on x86_64 and the doc comment holds — enabling FIPS mode without the provider panics rather than falling back:

| non-FIPS image + `[fips]` | result |
| --- | --- |
| `release-260812` | starts, `openssl_fips_enabled=true` |
| `release-260714` | panics, exit 101 — `fips.so: cannot open shared object file` |
| `release-260630` | block ignored, no `fips-support` line |

My values.yaml note was wrong, so `377ee6e` corrects it and rejects the flag at render time for official tags older than `release-260728`, following the Flux TTS precedent of failing before rollout instead of crash-looping. `release-260812` without `-fips` still renders since it works, and non-official tags are left alone so retagged private images don't false-fail.

BusyBox `-q`: yes, the input matched. `-q` returns 0 for `Connected`/`TrustBased` and 1 for `Disconnected`; `--quiet` returns 2 for all three, so the failure is option parsing, not a non-match.

`common/*.toml`: intentional scope, chart-only PR. Real gap for Compose deployments — follow-up PR, along with adding `helm template` to `ci.yml`, which is the gap that let your nil pointer through.

How it's tested: 13 render combinations across api/engine/licenseProxy/billing; all four TOMLs parsed with `tomllib`; probe commands and FIPS startup behavior run in the real `-fips` images rather than inferred; flag-off output byte-identical to your HEAD, so non-FIPS and pre-FIPS deployments are untouched. Your nil-pointer cases re-verified green on current HEAD.

jkroll-deepgram:
@therealevanhenry could I get another look? Your approval came in just before I pushed `377ee6e`, so that commit is unreviewed.

It's the one worth a second pair of eyes: it adds a hard render-time failure when `global.fips.enabled` is set on an official release tag older than `release-260728`. That blocks installs that previously rendered, which is intentional — those images have no OpenSSL FIPS provider and abort at startup with exit 101 rather than falling back — but it's a policy call, not just a fix. Non-official tags are deliberately left alone so retagged private images don't false-fail, and with the flag off nothing changes on any image.

Happy to loosen it to a warning-only path or narrow the tag matching if you'd rather.

PR #185, opened 2026-08-24
Render chart templates in CI

jkroll-deepgram:
Stacked on #184; base retargets to `main` when that merges.

## Why

`helm lint` parses the chart but never evaluates the templates, so render-time errors reach `main` uncaught. The nil pointer dereference on `.Values.global.fips.enabled` in #184 passed `helm lint` and both existing CI jobs, and would have broken `helm upgrade --reuse-values` for anyone upgrading from a chart version predating the key.

## What

A `helm template` step covering:

- Deployment shapes: defaults, License Proxy, FIPS on and off.
- A simulated `helm upgrade --reuse-values` from an older chart, with the newer value keys (`global.fips`, `fluxTts`, `api.features.speakV2*`) absent. This is the case that broke.
- Every `samples/*.values.yaml`, so the files we hand to customers are known to render.
- Negative cases for the configurations the chart deliberately rejects (FIPS on a pre-FIPS tag, `[fips]` in `customToml`, Flux TTS without a batch size, Flux TTS with Aura-2), so those guards cannot silently stop firing.

21 scenarios, roughly 35s.

## Verification

Confirmed it catches the bug that motivated it: reverting the nil guards makes the "newer keys absent" case fail with the original `nil pointer evaluating interface {}.enabled` error, and the step exits non-zero. Restored, all 17 pass. Also ran the script exactly as extracted from the workflow YAML rather than only in isolation.


jkroll-deepgram:
@therealevanhenry this CI improvement PR is ready for your re-review. Not urgent.

therealevanhenry (approved):
Approving. Ran the step as extracted from the YAML against this branch and against current `main`. 21/21 both ways. Reverting the nil guard fails "newer keys absent"; deleting the `maxBatchSize` guard fails "flux tts without batch." It catches what it claims to.

I pushed bcf9fec. "fips on a standard tag" used release-260714, which also predates the floor, so the floor check fired instead of the standard-image check. The negative cases now assert the error text they expect, and that case uses release-260812.

Two follow-ups, neither blocking:

- "newer keys absent" is a hand-maintained list. #186 already added four `logFormat` keys that aren't in it (they happen to be nil-safe). Diffing `values.yaml` leaf keys against the last chart tag and nulling the additions would keep it honest without anyone remembering.
- Samples 04, 06, and 08 only render with the CI overrides. That's by design for all three, but the step comment should say so. Sample 04's usage line also points at `04-aura-2-setup.yaml`; the file is `04-aura-2-setup.values.yaml`.

Description says 17 scenarios and ~35s. It's 21 and ~2s.

PR #186, opened 2026-08-24
Add logFormat values to select container log output format

jkroll-deepgram:
## Summary

Deepgram containers have supported four log output formats — `full`, `compact`, `pretty`, `json` — since [release-260319](https://developers.deepgram.com/changelog/2026/3/19), but the chart had no way to select one. `--log-format` is a CLI flag with no TOML or environment-variable equivalent, and the chart hardcoded each container's `args`, so neither `customToml` nor `extraEnv` could reach it.

This adds `logFormat` to all four components:

```yaml
api:          { logFormat: json }
engine:       { logFormat: json }
licenseProxy: { logFormat: json }
billing:      { logFormat: json }
```

Accepts `full`, `compact`, `pretty`, or `json`. An unrecognized value fails at template time rather than crash-looping the pod after rollout:

```
Error: api.logFormat must be one of full, compact, pretty, json (got "jsonl")
```

The default is empty, which renders args byte-identical to before — upgrades are a no-op unless the value is set.

## Flag placement

The flag is rendered **ahead of** the `serve` subcommand:

```
["-v","--log-format=json","serve","/etc/config/api.toml"]
["--log-format=json","serve","--config","/etc/config/billing.toml"]   # billing has no -v
```

This matters. `--log-format` is a top-level option on all four binaries, not a clap global — it appears in `<binary> --help` but not `<binary> serve --help`. Appending it makes the container exit immediately:

```
error: unexpected argument '--log-format' found
```

Note that the ["Kubernetes" example on the log formats docs page](https://developers.deepgram.com/docs/log-formats) currently shows the flag appended, via an `additionalArgs` value that has never existed in this chart. That example does not work for any of the four components and should be updated separately.

## Legacy JSON environment variable

The containers also accept a legacy `JSON` environment variable that forces JSON output, reachable today via `extraEnv`. Combining it with `--log-format` set to anything other than `json` does not resolve in favor of either setting — the process panics and the pod crash-loops:

```
thread 'main' panicked at src/logging.rs:348:13:
When --json arg or JSON env variable are set, cannot set --log-format to another format
```

The second commit rejects that combination at template time. The check is deliberately narrow, matching what actually panics:

| `extraEnv` | `logFormat` | Runtime | Chart |
|---|---|---|---|
| `JSON=true` | `json` | starts | allowed |
| `JSON=true` | `full` / `compact` / `pretty` | **panic** | **rejected** |
| `JSON=false` / `1` / `0` | any | starts | allowed |
| `JSON` via `valueFrom` | any | unknowable at render time | allowed |
| unset | any | starts | allowed |

Setting `JSON=true` with no `logFormat` is untouched, so anyone already using it as a workaround is unaffected.

## Testing

Verified against `release-260812` images on a GPU instance (NVIDIA L4). All four binaries behave identically:

| | Flag in `--help` | Absent from `serve --help` | Appended → fails | Prepended → works |
|---|---|---|---|---|
| API (`stem`) | ✅ | ✅ | ✅ | ✅ |
| Engine (`impeller`) | ✅ | ✅ | ✅ | ✅ |
| License Proxy (`hermes`) | ✅ | ✅ | ✅ | ✅ |
| Billing (`athena`) | ✅ | ✅ | ✅ | ✅ |

Engine output, past GPU init so this is real runtime behavior rather than arg parsing:

```
# logFormat: json
{"timestamp":"...","level":"INFO","fields":{"message":"GPU 0 passed the boot-time matmul check.","gpu":0},"target":"impeller"}

# default (full)
2026-08-24T21:26:10Z  INFO impeller: GPU 0 passed the boot-time matmul check. gpu=0
```

Template rendering verified for: unset (byte-identical to before), each of the four formats, an invalid value, every row of the `JSON` table above across all four components, and voice-agent mode where all three Engine deployments pick up the setting. `helm lint` and `helm-docs` are clean.

## Notes

- Left as an **Unreleased** changelog entry; no chart version bump.
- Setting `logFormat` on an image older than `release-260319` fails with `unexpected argument '--log-format' found` (confirmed against `release-260212`). This is not enforced by the chart — the version requirement is documented in `values.yaml`, and the failure is immediate and self-explanatory.
- `full` and `compact` produced byte-identical output in every sample captured, across all four binaries. They diverge only when span context exists, and these startup paths open no spans.
- `pretty` additionally emits source locations (`at src/config.rs:130`), which the docs' "color highlighting" description undersells.

### DIFF
--- a/charts/deepgram-self-hosted/CHANGELOG.md
+++ b/charts/deepgram-self-hosted/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## Unreleased
 
+## [0.45.0] - 2026-08-27
+
 ### Added
 
 - Added `api.logFormat`, `engine.logFormat`, `licenseProxy.logFormat`, and `billing.logFormat` to select a container's [log output format](https://developers.deepgram.com/docs/log-formats). Accepts `full`, `compact`, `pretty`, or `json`; an unrecognized value fails at template time rather than crash-looping the pod. Defaults to empty, which leaves each container on its own default (`full`) and renders the same arguments as before. Requires `release-260319` or later.
@@ -554,7 +556,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 - Initial implementation of the Helm chart.
 
-[unreleased]: https://github.com/deepgram/self-hosted-resources/compare/deepgram-self-hosted-0.43.0...HEAD
+[unreleased]: https://github.com/deepgram/self-hosted-resources/compare/deepgram-self-hosted-0.45.0...HEAD
+[0.45.0]: https://github.com/deepgram/self-hosted-resources/compare/deepgram-self-hosted-0.44.0...deepgram-self-hosted-0.45.0
+[0.44.0]: https://github.com/deepgram/self-hosted-resources/compare/deepgram-self-hosted-0.43.0...deepgram-self-hosted-0.44.0
 [0.43.0]: https://github.com/deepgram/self-hosted-resources/compare/deepgram-self-hosted-0.42.0...deepgram-self-hosted-0.43.0
 [0.42.0]: https://github.com/deepgram/self-hosted-resources/compare/deepgram-self-hosted-0.41.1...deepgram-self-hosted-0.42.0
 [0.41.1]: https://github.com/deepgram/self-hosted-resources/compare/deepgram-self-hosted-0.41.0...deepgram-self-hosted-0.41.1
--- a/charts/deepgram-self-hosted/CHANGELOG.md
+++ b/charts/deepgram-self-hosted/CHANGELOG.md
@@ -10,10 +10,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 - Added `api.logFormat`, `engine.logFormat`, `licenseProxy.logFormat`, and `billing.logFormat` to select a container's [log output format](https://developers.deepgram.com/docs/log-formats). Accepts `full`, `compact`, `pretty`, or `json`; an unrecognized value fails at template time rather than crash-looping the pod. Defaults to empty, which leaves each container on its own default (`full`) and renders the same arguments as before. Requires `release-260319` or later.
 
-### Notes
-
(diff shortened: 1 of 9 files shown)


jkroll-deepgram:
Rebased onto current main (`release-260826`) and re-verified end to end. Also bumped the chart to `0.45.0`

Merged main into this PR after yesterday's release.

New testing since last review:

- FIPS image variants (`release-260812-fips`, all four binaries). These didn't exist when the flag was first verified, and since `0.43.0` had to fix readiness probes for BusyBox, it seemed worth checking whether the FIPS base diverged here too. It doesn't — the flag is present in `--help`, absent from `serve --help`, appended fails, prepended honored, and the `JSON=true` panic reproduces identically. Engine verified on an L4.
- `release-260826` (the new default tag, stem 1.192.31). Same behavior; all four formats accepted.
- The default is provably inert. Rendered `main` and this branch from identical values and diffed the full manifest, rather than just eyeballing the args: lines. Eight lines differ, all cosmetic — `toJson` emits `["-v","serve",...]` where the hand-written YAML had `["-v", "serve", ...]`. Piping both through `kubectl apply --dry-run=client -o json` gives byte-identical output (98,334 bytes each), so no spurious rollouts and no GitOps drift for anyone who doesn't set the value.
- Every sample values file, rendered against both branch and `main`. Three fail identically on both (04, 06, 08 need volume config supplied) — pre-existing and unrelated.
- Voice-agent mode — all three Engine deployments plus API pick up the flag.

Taken together that covers all four binaries on both standard and FIPS images across two release tags, with the render matrix spanning default / each format / invalid value / JSON conflict / FIPS + logFormat / voice agent. The one thing I have not verified is the docs page's Docker Compose form under literal docker compose — I confirmed that argv fails via docker run on all four binaries, but never ran it through Compose itself. 

@therealevanhenry, thanks for catching the merge mangling in `660fe72`. Could you take a final look and approve if it reads right to you?

therealevanhenry (approved):
Re-verified on 0f9b8aa. helm lint is clean, all four components render the flag ahead of `serve` (billing without `-v`), an invalid value fails at render time, and the JSON=true + non-json combination is rejected with the right message. The 0.45.0 bookkeeping checks out: Chart.yaml bump, footer links repointed, missing 0.44.0 anchor added.

One non-blocking nit: the 0.45.0 entry is dated 2026-08-27. If you touch the branch again anyway, update it to the merge date. Not worth a push on its own.

PR #188, opened 2026-08-31
Add TTS first-byte latency dashboard and alert rules

jkroll-deepgram:
## What

Adds TTS first-byte latency monitoring to `monitoring/`. The existing `grafana_dashboard_template.json` covers STT, streaming, and capacity, but has no TTS panels.

| File | Contents |
|---|---|
| `grafana_tts_dashboard_template.json` | 5 panels: P50/P90/P99 first-byte latency, per-instance P99, raw vs transcoded means, request rate, mid-response failures |
| `prometheus_tts_alert_rules.yml` | 3 recording rules, 4 alerts |
| `prometheus_tts_alert_rules_test.yml` | promtool unit tests |
| `README.md` | TTS dashboard section, plus TTS alerts under Recommended Alerts |
| `tts_dashboard.png` | screenshot referenced from the README |

Built on the two histograms Engine already exposes: `engine_tts_first_transcoded_byte_latency` (caller-perceived, transcoding included) and `engine_tts_first_raw_byte_latency` (inference only). The gap between them separates a GPU regression from a transcode regression.

TTS first-byte latency dashboard

*Example data for illustration, not a benchmark.*

## Model coverage

| Model family | Endpoint | `tier` label |
|---|---|---|
| Aura / Aura-2 | `/v1/speak` | `aura`, `aura-2` |
| Flux TTS | `/v2/speak` | `flux` |

**Both are covered.** They share the same two first-byte histograms, distinguished by the `tier` label, and the panels don't filter on `tier` — so whichever model family the selected Engine serves is what you see. Flux TTS needs a dedicated Engine (it refuses to co-exist with Aura), so each is its own scrape target, selected via the **Scrape Job** dropdown.

Those two histograms are also the *entire* TTS latency surface for self-hosted builds. The per-pipeline internals (Aura-2 batcher timings, Flux TTS slot and queue depths, `time_to_first_code`) are registered only in Engine's hosted metric set, so there is nothing more granular available to chart.

Flux TTS requires a dedicated Engine and cannot share one with Aura models, so each is its own scrape target. The panels don't filter on `tier`, so they apply to whichever family the selected Engine serves.

Unrelated to `engine.flux` in the chart, which configures Flux streaming speech-to-text; its `engine_flux_*` metrics are not covered here.

Alert thresholds were baselined on Aura-2 — Flux TTS may have a different first-byte profile.

## Alerts

| Alert | Condition | Severity |
|---|---|---|
| `TTSLatencyHigh` | P90 > 200ms for 2m | warning |
| `TTSLatencyVeryHigh` | P90 > 400ms for 2m | critical |
| `TTSTailLatencyVeryHigh` | P99 > 500ms for 2m | critical |
| `TTSLatencyRegression` | P90 doubled vs 1h earlier, for 15m | warning |

Every threshold lands on an actual histogram bucket boundary, so each condition is exact rather than dependent on `histogram_quantile` interpolating inside a bucket. All four are gated on a minimum request rate, since a quantile over a near-empty window just tracks the single slowest request. Thresholds are documented as a starting point to baseline against, not as an SLO.

## Validation

- `promtool check rules` — 7 rules, passes
- `promtool test rules` — 3 scenarios (healthy / warning / both criticals), passes
- Rules loaded into a live Prometheus scraping a real Engine; all 7 evaluate with `health=ok`
- Dashboard imports and renders on Grafana 12.0.0; all panel queries verified through the datasource proxy
- Alert firing verified end to end, including the rendered annotation

## Notes for review

- **Separate dashboard rather than new panels in the existing template.** Avoids conflicts with the release automation that updates that file, and avoids mixing `job`-based filters into a `namespace`-based dashboard. Happy to merge them instead if preferred.
- **Panels filter on `job`, not `namespace`.** Keeps the dashboard usable for Docker Compose as well as Kubernetes. The README notes how to add a `namespace` filter.
- **README states Grafana 12.0.0 for this dashboard** vs 12.3.0 for the existing one. Deliberate — this dashboard is schemaVersion 41 and uses only long-stable panel features, so it imports on 12.0. Reads as inconsistent, so flagging it.
- **TTS metrics are registered lazily.** An Engine that hasn't served a TTS request exposes no `engine_tts_*` series, so it won't appear in the Scrape Job dropdown until it does. Documented in the README; detecting a silent Engine needs a separate `up`-based check.


pcgeek86 (approved):

monitoring/prometheus_tts_alert_rules.yml
```diff
@@ -0,0 +1,48 @@
...
+      - alert: TTSLatencyHigh
+        expr: tts:ttfb_p90:5m > 0.2 and tts:rps:5m > 0.1
+        for: 5m
+        labels: { severity: warning }
+        annotations:
+          summary: "{{ $labels.job }} P90 TTFB {{ $value | humanizeDuration }} (>200ms)"
+
+      - alert: TTSLatencyVeryHigh
+        expr: tts:ttfb_p90:5m > 0.4 and tts:rps:5m > 0.1
+        for: 2m
```
leah-deepgram, line 28: Hmm maybe this LatencyVeryHigh should have the same `for: 5m` as the above LatencyHigh alert. Currently, if p90 goes above 400ms, at t+2 min, it'll fire this critical alert first, then at t+5 min, it'll start firing the warning alert, which seems like the wrong order? And the warning alert is not suppressed despite a critical alert has already fired. I think the threshold (200ms vs 400ms) should be sufficient for the severity. If we do want an earlier alert, maybe we can have for: 2m for the warning, and then for: 5m for the critical.

jkroll-deepgram: Good catch — fixed in 28e43d8. You're right that the ordering was backwards.

I went with the same `for: 2m` on both P90 alerts rather than warning `2m` / critical `5m`, because making the critical slower than the warning means a severe regression pages *later* than a mild one. With both at `2m` the threshold carries the severity, as you suggested, and they can't fire out of order. `TTSTailLatencyVeryHigh` was already `2m`, so all three are now consistent; `TTSLatencyRegression` stays at `15m` since it's a slow-drift signal rather than a threshold breach.

On the warning not being suppressed — that one can't be solved in the rule file, since Prometheus has no notion of one alert superseding another. It needs an Alertmanager inhibit rule, so I've documented this in the README:

```yaml
inhibit_rules:
  - source_matchers: [severity="critical"]
    target_matchers: [severity="warning"]
    equal: ['job']
```

`equal` is on `job` only, not `alertname` — the alert names differ, so matching on those would never inhibit anything.

Worth noting the recording rules use a `[5m]` rate window, so a step change takes a few minutes to fully register regardless of `for:`. Real detection lands around 3–5 minutes either way; `for: 2m` mainly guards against a single scrape blipping the quantile.

jkroll-deepgram:
@pcgeek86 I confirmed that Aura-2 and Flux TTS publish the same metrics, so this template is applicable for both 🎉 I clarified that point in the wording for these templates/readme, as well as in a separate internal PR for our [metrics guide](https://developers.deepgram.com/docs/metrics-guide) doc.

leah-deepgram (approved):

PR #190, opened 2026-09-17
Fix License Proxy bypass when Billing is also enabled

pcgeek86:
**Short description**: If license proxy is enabled, always configure API and Engine to use that URL for license server endpoint. Otherwise, if the license proxy is not enabled, they can use the billing API directly.

## Summary
- API and Engine `[license]` config picked `billing.enabled` over `licenseProxy.enabled`, so enabling both together caused API/Engine to connect straight to the Billing container, silently bypassing a fully-deployed and correctly-configured License Proxy.
- The License Proxy's own config (`license-proxy.config.yaml`) already chains to Billing when `billing.enabled` is true — this fix makes that path actually reachable by routing API/Engine through the proxy whenever it's enabled, regardless of the Billing setting.
- The `keepUpstreamServerAsBackup` fallback to `https://license.deepgram.com` is now only added when Billing is not enabled, since that public endpoint is unreachable in airgapped deployments.
- Added a CHANGELOG entry under Unreleased; no version bump since this is not ready to merge yet.

## Test plan
- [x] `helm template` rendered `api.config.yaml` / `engine.config.yaml` for all four combinations of `licenseProxy.enabled` / `billing.enabled` and confirmed `[license].server_url` matches the intended behavior (proxy-only, billing-only, and neither unchanged; both-enabled now routes through the proxy).
- [ ] Manual validation against a live airgapped cluster with both License Proxy and Billing enabled, confirming Engine/API successfully license through the proxy chain.


jeniyat (approved):
LGTM

therealevanhenry (approved):

